PR #21, opened 2024-07-28
optional filemanager

paiyayay:
Current configuration of dashboard file paths takes too much time.

There should be easier way to change the file manager.
I recommend the env var solution as the easiest solution, though the users have to set the env var themselves.

There can be a text prompt right before dashboard file paths config, which asks users their file managers. This text prompt then sets the $FILEMANAGER to the given application.

Testing if given application is available can be good to check.

### DIFF
--- a/options.ts
+++ b/options.ts
@@ -668,29 +668,29 @@ const options = mkOptions(OPTIONS, {
                 left: {
                     directory1: {
                         label: opt("󰉍 Downloads"),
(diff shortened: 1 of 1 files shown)


jgroboredo:
The text prompt would probably be a good idea, also following the fact that already exists one for the terminal (and would, therefore, be a consistent approach).

Another option could be to use `xdg-open` or `handlr` or something similar.

PR #53, opened 2024-07-31
added charging detection to battery icon

matavach:
Simple change to dynamically change icon if battery is actively charging.


Jas-SinghFSU:
Thank you for the contribution.

PR #60, opened 2024-08-02
Added an OSD for microphone inputs.

### LINKED ISSUE #58: OSD for mic mute
volume mute has OSD but mic mute does not have it. It would be great to notified when pressed mute.

Jas-SinghFSU:
The OSD now displays itself when Input volume is changed or mute is toggled.0

closes #58


greeid:
Any way to display this in the panel?

PR #64, opened 2024-08-02
added 'increment' option for Inputter

### LINKED ISSUE #63: Font weight breaks on spinner button press
**Describe the bug**
Font weight (and possibly other settings) break on spinner button press. This is because font weight only accepts increments of 100, but the spinner buttons change the value by 1. This causes all CSS to error.

**To Reproduce**
Any press on spinner button for "Font Weight"

**Expected behavior**
Spinner buttons to adjust values in valid increments

**Screenshots**

 - Distribution:  Arch Linux
 - Window Manager/Desktop Environment:  Hyprland

matavach:
fixes #63 

Added an increment option for Inputter. This will allow us to set an scaling increment factor for each spinner button on the config panel.

I also noticed that font-weight will need a minimum value of 100, so I added that as well.

I haven't tested all spinner numbers yet to confirm that font-weight is the only option affected by the issue, so I will leave the issue open until I test them all and commit any more changes related to it. 

Edit: Font weight also has a max value of 900, so I added that as well.

### DIFF
--- a/lib/types/options.d.ts
+++ b/lib/types/options.d.ts
@@ -22,6 +22,7 @@ export type RowProps<T> = {
     max?: number
     min?: number
     subtitle?: string
+    increment?: number
 }
 
 export type OSDOrientation = "horizontal" | "vertical";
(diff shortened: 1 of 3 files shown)


Jas-SinghFSU:
Perfect.

PR #71, opened 2024-08-03
added null check for weather key

### LINKED ISSUE #70: Weather module does not check for null API key
**Describe the bug**
Weather module does not check for a null key. Results in widget constantly trying to parse a null value and throwing errors.

**Expected behavior**
Weather module does not load on null value

 - Distribution: Arch

matavach:
fixes #70 

I see two approaches for solving this, null checking on the weather module itself or null checking in the index.ts for the calendar module.

I chose the former, since it would keep the fix if we decide to use the weather module somewhere else in the code.

I added a visible property to the top level box, because otherwise it would render an empty box.

image

with the visible property also doing a null check:
image

Do we want to include a warning for users if weather is disabled?


Jas-SinghFSU:
Hmmm, is the calendar not displaying default values? My original idea was to leave it in there and display 0s for the weather values until someone put in the weatherAPI key. That way, the user would know that the Weather module exists upon first glance and they can activate it with a key.

matavach:
It does display zeroes, however, as is, it will poll constantly and throw errors nonstop.
I will adjust this to display the 0s without polling.

matavach:
With the new commit, this is the view with no api key available.
image
It also stops polling until a key is presented.

modules/menus/calendar/weather/index.ts
```diff
@@ -19,6 +19,9 @@ const WeatherWidget = () => {
                 Utils.merge(
                     [key.bind("value"), interval.bind("value"), location.bind("value")],
                     (weatherKey, weatherInterval, loc) => {
+                        if(!weatherKey){
```
Jas-SinghFSU, line 22:
```suggestion
                        if(!weatherKey) {
```

Issue #74, opened 2024-08-04
Nix build failure caused by GdkPixbuf-2.0


mitsuruu:
**Describe the bug**

**To Reproduce**
- Clone the repository
- Run `nix build`

```
error: builder for '/nix/store/bjrkxrfk0z3q4gwnidhm4ilz7dyimygh-hyprpanel-latest.drv' failed with exit code 1;
       last 10 log lines:
       > Running phase: patchPhase
       > Running phase: updateAutotoolsGnuConfigScriptsPhase
       > Running phase: configurePhase
       > calling 'preConfigure' function hook '_multioutConfig'
       > no configure script, doing nothing
       > Running phase: buildPhase
       > 4 | import GdkPixbuf from "types/@girs/gdkpixbuf-2.0/gdkpixbuf-2.0.js";
       >                           ^
       > error: Could not resolve: "types/@girs/gdkpixbuf-2.0/gdkpixbuf-2.0.js". Maybe you need to "bun install"?
       >     at /build/nrn2j05b1w5qinz0a77jva3xqj135kcc-source/modules/menus/dashboard/profile/index.ts:4:23
       For full logs, run 'nix log /nix/store/bjrkxrfk0z3q4gwnidhm4ilz7dyimygh-hyprpanel-latest.drv'.
error: 1 dependencies of derivation '/nix/store/0r8shgja2cgq0d3sgbmic1fi1vpb5kg8-hyprpanel.drv' failed to build
```

**Expected behavior**
HyprPanel should build

**Screenshots**

**Desktop (please complete the following information):**
 - Distribution: NixOS
 - Window Manager/Desktop Environment: Hyprland

**Additional context**
This issue is caused by `modules/menus/dashboard/profile/index.ts:4` with
```ts
import GdkPixbuf from "types/@girs/gdkpixbuf-2.0/gdkpixbuf-2.0.js";
```
which (as far as I'm aware anyway) isn't part of the default Ags types. Removing this line, and removing line 30 resolves the build issue, and doesn't seem to cause any issues in my testing.

mitsuruu:
This issue should be closed with when the `pixbuf_fix` branch gets merged to `master`.

PR #75, opened 2024-08-04
Resolve nix build failures (#74)

mitsuruu:
Addresses issue #74, however may have some unintended side-effects.

Another possible solution would be to include `@girs/gdkpixbuf-2.0` in source, or package it through Nix as a dependency, however I haven't tried that as of yet.

### DIFF
--- a/modules/menus/dashboard/profile/index.ts
+++ b/modules/menus/dashboard/profile/index.ts
@@ -1,7 +1,6 @@
 import icons from "../../../icons/index.js";
 import powermenu from "../../power/helpers/actions.js";
 import { PowerOptions } from "lib/types/options.js";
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Could you checkout out branch `pixbuf_fix` and see if the issue still persists? It's used for checking if the selected profile image is a valid image and if not it will fallback to the default icon.

I've replaced the import and hopefully you shouldn't get build failures.

mitsuruu:
Just pulled down the `pixbuf_fix` branch and I can confirm that the issue no longer persists. Seems like replacing the import has resolved the issue of building and running under Nix.

I'll go ahead and close this PR.

PR #88, opened 2024-08-06
added options for bar media module

### LINKED ISSUE #87: Show track artist on the bar media module.
Currently, the media bar only shows the track title.
User's should be able to optionally show the track artist as well.

matavach:
Added 3 options for the bar media module
- Truncate
- Truncate Size
- Show Artist

fixes #87

### DIFF
--- a/modules/bar/media/index.ts
+++ b/modules/bar/media/index.ts
@@ -1,6 +1,10 @@
 import Gdk from 'gi://Gdk?version=3.0';
 const mpris = await Service.import("mpris");
 import { openMenu } from "../utils.js";
+import options from "options";
+import media from 'modules/menus/media/index.js';
+
+const { show_artist, truncation, truncation_size } = options.bar.media;
 
 const Media = () => {
     const activePlayer = Variable(mpris.players[0]);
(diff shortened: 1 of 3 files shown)


modules/bar/media/index.ts
```diff
@@ -76,12 +83,17 @@ const Media = () => {
...
                         }),
-                        Widget.Label({
+                        Widget.Box({
                             class_name: "bar-button-label media",
-                            label,
-                            truncate: "end",
-                            wrap: true,
-                            maxWidthChars: 30,
+                            child: Utils.merge(
+                                [truncation.bind("value"), truncation_size.bind("value")],
```
Jas-SinghFSU, line 89: Should we move these bindings to the mediaLabel so they're all consolidated? Then we can handle the truncation in that `Utils.watch` as well. That way we don't need to nest our Widget.Label in a Widget.Box and we can just call the label as:
```
Widget.Label({
  label: mediaLabel
}))
```

widget/settings/pages/config/bar/index.ts
```diff
@@ -72,11 +72,14 @@ export const BarSettings = () => {
 
                 Header('Media'),
                 Option({ opt: options.theme.bar.buttons.media.spacing, title: 'Inner Spacing', subtitle: 'Spacing between the icon and the label inside the buttons.', type: 'string' }),
+                Option({ opt: options.bar.media.show_artist, title: 'Show Track Artist', type: 'boolean' }),
+                Option({ opt: options.bar.media.truncation, title: 'Truncate Media Label', type: 'boolean' }),
+                Option({ opt: options.bar.media.truncation_size, title: 'Truncation Size', type: 'number' }),
```
Jas-SinghFSU, line 77: Should we set a minimum truncation size here?

modules/bar/media/index.ts
```diff
@@ -1,6 +1,10 @@
 import Gdk from 'gi://Gdk?version=3.0';
 const mpris = await Service.import("mpris");
 import { openMenu } from "../utils.js";
+import options from "options";
+import media from 'modules/menus/media/index.js';
```
Jas-SinghFSU, line 5:
```suggestion
```
Seems like we're not using `media` anywhere.

widget/settings/pages/config/bar/index.ts
matavach: I intentionally didn't add one, but I will.

PR #112, opened 2024-08-12
Workspaces are now handled properly for scrolling when there are no rules defined for them.

### LINKED ISSUE #110: Scroll wheel does not work on workspace buttons when workspace rules are not set.
**Describe the bug**
Scroll wheel does not work on workspace buttons when workspace rules are not set. Even if some of them are set, the scroll wheel will switch to those workspaces only.

**To Reproduce**
Unset the all workspace rules (e.g. `workspace = 1, monitor:DP-2, default:true`). Try to use scroll wheel on W/S buttons in the bar.

**Expected behavior**
scroll wheel function should work irrespective of workspace rules are set or not.

 - Distribution: Arch Linux

Jas-SinghFSU:
If no rules were defined at all for workspaces then scrolling would break. Scrolling now works appropriately when no rules are defined.

closes #110

### DIFF
--- a/modules/bar/workspaces/index.ts
+++ b/modules/bar/workspaces/index.ts
@@ -67,25 +67,43 @@ const Workspaces = (monitor = -1, ws = 8) => {
     })
 
     const goToNextWS = (): void => {
-        const curWorkspace = hyprland.active.workspace.id;
-        const indexOfWs = currentMonitorWorkspaces.value.indexOf(curWorkspace);
-        let nextIndex = indexOfWs + 1;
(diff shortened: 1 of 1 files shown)


Shubham-550:
Scrolling of workspace stops on the other monitor at the workspace which has a window.

**To Reproduce** 
no workspace rules enabled
Monitor specific - off

Jas-SinghFSU:
@Shubham-550 This is because hyprland automatically moves your cursor to the workspace that you switch to, which causes you to lose focus of the HyprPanel workspace module.

This isn't something HyprPanel can control and would have to be configured in the hyprland.conf if a settings exists.

PR #122, opened 2024-08-13
Make media label togglable

orxngc:
Adds the option to completely hide the song title if so desired.
Some people may only want the media icon + the mpris control widget, not the title..

### DIFF
--- a/modules/bar/media/index.ts
+++ b/modules/bar/media/index.ts
@@ -51,7 +51,7 @@ const Media = () => {
                     : `${truncatedLabel.substring(0, truncatedLabel.length - 3)}...`;
         } else {
             songIcon.value = getIconForPlayer(activePlayer.value?.identity || "");
-            return `${songIcon.value} Media`;
+            return `Media`;
         }
     });
 
--- a/modules/bar/media/index.ts
+++ b/modules/bar/media/index.ts
@@ -4,7 +4,7 @@ import { openMenu } from "../utils.js";
 import options from "options";
 import { getCurrentPlayer } from 'lib/shared/media.js';
 
-const { show_artist, truncation, truncation_size } = options.bar.media;
+const { show_artist, truncation, truncation_size, show_label } = options.bar.media;
 
 const Media = () => {
     const activePlayer = Variable(mpris.players[0]);
@@ -33,8 +33,8 @@ const Media = () => {
 
     const songIcon = Variable("");
 
-    const mediaLabel = Utils.watch("󰎇 Media 󰎇", [mpris, show_artist, truncation, truncation_size], () => {
-        if (activePlayer.value) {
+    const mediaLabel = Utils.watch("󰎇 Media 󰎇", [mpris, show_artist, truncation, truncation_size, show_label], () => {
+        if (activePlayer.value && show_label.value) {
             const { track_title, identity, track_artists } = activePlayer.value;
             songIcon.value = getIconForPlayer(identity);
             const trackArtist = show_artist.value
@@ -50,8 +50,8 @@ const Media = () => {
                     ? `${truncatedLabel}`
                     : `${truncatedLabel.substring(0, truncatedLabel.length - 3)}...`;
         } else {
-            songIcon.value = "";
-            return "󰎇 Media 󰎇";
+            songIcon.value = getIconForPlayer(activePlayer.value?.identity || "");
+            return `${songIcon.value} Media`;
         }
     });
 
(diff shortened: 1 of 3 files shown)


painerp:
I think you changed to default value to 0, it is probably better to change the min value of the option instead.
https://github.com/Jas-SinghFSU/HyprPanel/blob/ea2ebc5c4b64ba4eeb6dfe773d5e1482ebdf587a/widget/settings/pages/config/bar/index.ts#L77

orxngc:
O, I see, I'm an idiot. Justamoment.

Jas-SinghFSU:
I can definitely see this being useful but rather than truncating to 0, I'd rather just toggle the value off. Reason for that is if you leave an empty label, you have a side-effect that creates empty padding to the right side of the icon. This'll create an off-centered icon as a result.

orxngc:
Ah good point, I'll change that

orxngc:
Alright I made it togglable.. it works.
image

widget/settings/pages/config/bar/index.ts
```diff
@@ -73,13 +73,14 @@ export const BarSettings = () => {
                 Header('Media'),
                 Option({ opt: options.theme.bar.buttons.media.spacing, title: 'Inner Spacing', subtitle: 'Spacing between the icon and the label inside the buttons.', type: 'string' }),
                 Option({ opt: options.bar.media.show_artist, title: 'Show Track Artist', type: 'boolean' }),
-                Option({ opt: options.bar.media.truncation, title: 'Truncate Media Label', type: 'boolean' }),
-                Option({ opt: options.bar.media.truncation_size, title: 'Truncation Size', type: 'number', min: 10 }),
+                Option({ opt: options.bar.media.show_label, title: 'Toggle Media Label', type: 'boolean' }),
+                Option({ opt: options.bar.media.truncation, title: 'Truncate Media Label', subtitle: 'Only applicable if Toggle Media Label is enabled', type: 'boolean' }),
+                Option({ opt: options.bar.media.truncation_size, title: 'Truncation Size', subtitle: 'Only applicable if Toggle Media Label is enabled', type: 'number', min: 0 }),
```
Jas-SinghFSU, line 78:
```suggestion
                Option({ opt: options.bar.media.truncation_size, title: 'Truncation Size', subtitle: 'Only applicable if Toggle Media Label is enabled', type: 'number', min: 10 }),
```
Let's keep this at 10 since now we have a toggle.

modules/bar/media/index.ts
```diff
@@ -50,8 +50,8 @@ const Media = () => {
                     ? `${truncatedLabel}`
                     : `${truncatedLabel.substring(0, truncatedLabel.length - 3)}...`;
         } else {
-            songIcon.value = "";
-            return "󰎇 Media 󰎇";
+            songIcon.value = getIconForPlayer(activePlayer.value?.identity || "");
+            return `${songIcon.value} Media`;
```
Jas-SinghFSU, line 54:
```suggestion
            return `Media`;
```
Let's leave it to the icon label to handle icons. That way we have the flexibility to assigning styles to icons and configuring spacing between the icon and the label.

modules/bar/media/index.ts
```diff
@@ -33,8 +33,8 @@ const Media = () => {
 
     const songIcon = Variable("");
 
-    const mediaLabel = Utils.watch("󰎇 Media 󰎇", [mpris, show_artist, truncation, truncation_size], () => {
-        if (activePlayer.value) {
+    const mediaLabel = Utils.watch("󰎇 Media 󰎇", [mpris, show_artist, truncation, truncation_size, show_label], () => {
```
Jas-SinghFSU, line 36:
```suggestion
    const mediaLabel = Utils.watch("Media", [mpris, show_artist, truncation, truncation_size, show_label], () => {
```
As mentioned below, we'll leave this to the widget that draws the icons rather than drawing it in the label itself. I should have done it this way initially so let's fix it while we're at it.

modules/bar/media/index.ts
```diff
@@ -65,6 +65,7 @@ const Media = () => {
                         Widget.Label({
                             class_name: "bar-button-icon media",
                             label: songIcon.bind("value"),
+                            visible: show_label.bind("value"),
```
Jas-SinghFSU, line 68:
```suggestion
                            label: songIcon.bind("value").as(v => v || "󰝚"),
```
We'll handle drawing the music note icon here if the player doesn't generate an icon itself (fallback icon). And since this Widget.Label handles the icon at all times, we don't need to toggle visibility.

orxngc:
Makes sense, I applied the changes.

Jas-SinghFSU:
Looks good, merging... Thanks for the contribution ❤️.

PR #126, opened 2024-08-14
DropDownMenu: Ensure they stay within screen bounds at scaling factors

ElectroPerf:
- Adjusted monitor width and height calculations to properly account for Hyprland and GDK scaling factors.
- Added handling for vertical monitor transformations by swapping width and height.
- Implemented boundary checks to keep dropdown menus within the visible screen area.
- Updated `moveBoxToCursor` function to dynamically adjust margins based on screen dimensions and scaling.

This fix resolves the issue where dropdown menus would go out of the screen when using a scaling factor of 1.2 or higher.

Change-Id: I5298c5d44dc0204c2676fc4aefd03a05caf30d5a

### DIFF
--- a/modules/menus/DropdownMenu.ts
+++ b/modules/menus/DropdownMenu.ts
@@ -13,7 +13,7 @@ export const Padding = (name: string) =>
         setup: (w) => w.on("button-press-event", () => App.toggleWindow(name)),
     });
 
-const moveBoxToCursor = (self: any, fixed: boolean) => {
+const moveBoxToCursor = (self, fixed) => {
     if (fixed) {
         return;
     }
@@ -27,7 +27,7 @@ const moveBoxToCursor = (self: any, fixed: boolean) => {
             const monitorInfo = await bash('hyprctl monitors -j');
             const parsedMonitorInfo = JSON.parse(monitorInfo);
 
-            const foundMonitor = parsedMonitorInfo.find((monitor: Monitor) =>
+            const foundMonitor = parsedMonitorInfo.find((monitor) =>
                 monitor.id === hyprland.active.monitor.id
             );
             hyprScaling = foundMonitor?.scale || 1;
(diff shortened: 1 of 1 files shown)


ElectroPerf:
fixes this 
screenshot_14082024_214834

and makes it look like
image

Jas-SinghFSU:
Not sure I understand what this PR is for; this issue was fixed before with hyprland scaling being accounted for.

ElectroPerf:
> Not sure I understand what this PR is for; this issue was fixed before with hyprland scaling being accounted for.

But it still exists tho, as evident from the screenshot

zoran-olujic-cl:
I can confirm that this issue still exists.

ElectroPerf:
Yep

Jas-SinghFSU:
Hmm, can there be scenarios where both gdk AND hyprland scaling can exist in parallel? I see that the conditional statement was removed so that implies that hyprland scaling is always taken into account. I guess it make sense to default to it.

ElectroPerf:
Definitely makes sense to default to it. But about parallel, naah idts

Jas-SinghFSU:
This ones tricky because you can have Hyprland default to 2x scale for hiDPI displays which, if we divide by that scale by default (removing the else), in cases where there's GDK scaling, it will once again mess up the menu positioning.

So imagine this config provided by gdk and hyprland:
Hyprland Scale: 2x
Gdk Scale: 2x
GdkDPIScale: 0.75x

This will cause the menus to spawn half a screen away because we're dividing the menu margins by half. There is no clear indicator of why Hyprland auto applies 2x scaling on hiDpi displays even if I set the config to 1x...

The only thing I can think of in this case is providing presets that account for:
GDK_SCALING + GDK_DPI_SCALING + Hyprland Scaling
GDK_SCALING + GDK_DPI_SCALING 
Hyprland Scaling

And you can toggle between them depending on your display.

Jas-SinghFSU:
More specifically, set your resolution to 
monitor=DP-1,3840x1600@144,1440x350,1.5 # <--- this is the scale

env = GDK_SCALE,2
env = GDK_DPI_SCALE,0.5
env = XCURSOR_SIZE,15
env = QT_AUTO_SCREEN_SCALE_FACTOR,1

The above ^ for gdk scale. Now with your config, realize that the positioning of the menus is messed up.

Jas-SinghFSU:
Added a configuration option for the various scaling scenarios in `Configuration General > Scaling > Scaling Priority`. Could you pull master and see if selecting the right one - probably `hyprland` or `both` in your case - fixes the issue for you guys?

Jas-SinghFSU:
Closing... refer to #277

PR #127, opened 2024-08-14
Fixes inconsistent behavior when controlling multiple devices with brightnessctl

painerp:
Fixes a small issue when multiple devices are controlled by brightnessctl and the current and max values are coming from Device 1 but changes are applied to Device 2.

This should probably be handled by addressing all devices at once and changing them to the desired percentages.
For example by changing all devices with the backlight class: `brightnessctl -c backlight s 20%`.
This sadly seems to be broken in brightnessctl v5.0 though.


Jas-SinghFSU:
Looks good, thanks!

PR #134, opened 2024-08-16
Fixes OSD muting inconsistencies 

painerp:
When muting via command, the previous values were still visible in the level bar and label.

The `is_muted !== false` looks counterintuitive, but seems to me like the best way to handle the null return (e.g. when no microphone is available) which should also be treated as a muted state.


Jas-SinghFSU:
Is this pull request to show the bar and the counter at 0 whenever muting?

painerp:
Exactly, yes it will show a level of 0 in both the bar and label whenever muting, and the current percentage when unmuting.


_microphone volume is at 50%  (25% was my speaker volume changed before)_ 
Before:
before

After:
after

Jas-SinghFSU:
@painerp Something interesting I found is that there's no standard UX design for OSD when it comes to muting. Windows displays the volume number and bar at the current values even when muted (similar to hyprpanel's current behvaior) and macOS behaves the way you've implemented it.

I still think this is a good implementation but I would rather make this optional and opt-in rather than changing it entirely; since both behaviors are 'normal'. If you would like to implement these options yourself, that'd be awesome. But no worries if not, I can create a branch based off of yours that you can merge into yours.

painerp:
Ah shit I fucked up, to tiered to fix it now. I will just open a new one :)

PR #156, opened 2024-08-19
implemented xdg-open and '$TERMINAL' environment variable

matavach:
Removed explicit references to 'dolphin' and replaced them with `xdg-open`

Also removed the hyprpanel option for setting the terminal, and instead replaced references to that with the `$TERMINAL` environment variable.
  - note: as far as I can tell, XDG doesn't have a simple way of opening terminal applications, but the standard workaround for this is to use the `$TERMINAL` envar.

Implementing this PR will require us to add `xdg-desktop-portal` as a dependency (and maybe  `xdg-desktop-portal-hyprland` as well) and either set the `$TERMINAL` envar during the installation script, or include instructions for users to set it.

fixes #21

### DIFF
--- a/modules/menus/dashboard/stats/index.ts
+++ b/modules/menus/dashboard/stats/index.ts
@@ -2,7 +2,6 @@ import options from "options";
 import { GPU_Stat } from "lib/types/gpustat";
 import { dependencies } from "lib/utils";
 
-const { terminal } = options;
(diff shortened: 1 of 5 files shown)


orxngc:
this has merge conflicts

rubiin (approved):

rubiin:
@matavach can you fix the conflicts?

PR #163, opened 2024-08-21
Implement Cava module (WIP)

matavach:
* Adding Cava module that will display and provide configuration for Cava.

Very early stages of work in progress.


Jas-SinghFSU:
Closing because stale.

PR #169, opened 2024-08-22
Adds more window title options

painerp:
Adds multiple new Window Title options:
- Icon Toggle for Custom Titles
- Custom Window Title Toggle:
  - If disabled, the title will default to either the client-provided title or, if that’s empty, the class name.
- Truncation Options for the new Title
  - Added a toggle for truncating the new title.
  - Added options to set the truncation length.


Jas-SinghFSU:
Noticed the following issues:
1. Show Custom Title hides the icon (even with icon enabled)
2. Truncate window title doesn't seem to work (enabled or disabled, no matter the truncation size)

painerp:
Oh my bad, I might have broken something while merging your changes. I will look at them in a few days when I find some more time.

Jas-SinghFSU:
No worries :+1:

painerp:
@Jas-SinghFSU Alright I finally got some time on my hands again and checked your issues.
1. Added custom icon support even when the custom title is disabled.
2. Applied truncation to custom titles as well.

I hope I understood your points correctly, else just give me heads up and i will fix it. Thanks!

modules/bar/window_title/index.ts
```diff
@@ -132,28 +133,55 @@ const filterTitle = (windowtitle: ActiveClient) => {
     };
 };
 
+const defaultTitle = (client: ActiveClient) => {
+    let title = client.title;
+    // If the title is empty or only filled with spaces, fallback to the class name
+    if (title.length === 0 || title.match(/^ *$/)) {
+        title = client.class;
+    }
+    return title;
```
Jas-SinghFSU, line 142:
```suggestion
    let title = client.title;
    // If the title is empty or only filled with spaces, fallback to the class name
    if (title.length === 0 || title.match(/^ *$/)) {
        title = client.class;
    }
    return title;
```
We return class by default for windows. I don't necessarily dislike having the title as well but if we're going to display a title we should make that an option instead.

Everything else looks great :+1:

Jas-SinghFSU (changes requested):

Jas-SinghFSU (approved):
Lookin' juicy :+1:

Jas-SinghFSU:
As always, appreciate the contribution :heart:

PR #192, opened 2024-08-26
Made unoccupied workspaces setting more compaitble with # of workspaces assigned.

### LINKED ISSUE #190: [BUG] Workspace widget rework (from #189)
**Describe the bug**
Bar after pull #189 still doesn't have parity with waybar functionality. A few notables:
- When on startup, only 1 workspace is shown when I would expect the number of workspaces to be whatever I set the value of `Total Workspaces` to be in dashboard menu (these would be "persistent" workspaces IF the `hide unoccupied` workspaces option is selected). In other words, _I would like workspaces 1, 2, 3, 4 to be shown persistently if I have set my `Total Workspaces` number to 4._ Additional workspaces (>5) should still follow the "hide unoccupied" rules.
##### `Total Workspaces` is set to 4, yet only 2 workspaces shown (only 2 are active):
2024-08-26T11:50:39,610504185+10:00

- When I do have many workspaces open, when I (for example) here have 5 workspaces open with nothing in workspace 3, pressing `$mainMod+shift+3` when on workspace 6, I expect the window to be moved to workspace 3, which will then open inbetween workspaces 2 and 4. However, the behaviour simply keeps it at "index" 5, but "number" 3.
##### Initial state
2024-08-26T11:39:30,992117013+10:00
##### State after $mainMod+shift+3
2024-08-26T11:39:39,419665199+10:00

- OPTIONAL: As a follow on from this behaviour, when I log into hyprland, only 1 workspace is shown (no. 1). Pressing keybindings to go to workspace 8 will then only show workspace 8 instead of both workspace 1 and 8 (IMPORTANT: This is assuming that workspace 1 is "persistent", which IMO should only be the case only when the `Total Workspaces` is set to 1. Maybe setting this value to zero would the avoid any persistent workspaces again, just like the current behaviour?
2024-08-26T11:31:22,164784057+10:00
TL;DR: I think that the `Total Workspaces` field should be set to function as `Persistent Workspaces` *if and only if* the `Hide Unoccupied` field is selected.

Arch Linux lastest, hyprland

**Additional Info**
Sorry for the long report oof. In all honesty though this is an awesome improvement in adding waybar's functionality, I appreciate it!

Jas-SinghFSU:
Make Hyprpanel's occupied workspace behavior consistent with # of workspaces assigned.
closes #190

### DIFF
--- a/modules/bar/workspaces/helpers.ts
+++ b/modules/bar/workspaces/helpers.ts
@@ -38,7 +38,10 @@ export const getWorkspaceRules = (): WorkspaceMap => {
 
         JSON.parse(rules).forEach((rule: WorkspaceRule, index: number) => {
             if (Object.hasOwnProperty.call(workspaceRules, rule.monitor)) {
-                workspaceRules[rule.monitor].push(index + 1);
+                const workspaceNum = parseInt(rule.workspaceString, 10);
+                if (!isNaN(workspaceNum)) {
+                    workspaceRules[rule.monitor].push(workspaceNum);
+                }
             } else {
                 workspaceRules[rule.monitor] = [index + 1];
             }
--- a/modules/bar/workspaces/helpers.ts
+++ b/modules/bar/workspaces/helpers.ts
@@ -16,7 +16,10 @@ export const getWorkspacesForMonitor = (curWs: number, wsRules: WorkspaceMap, mo
     }
 
     const monitorMap = {};
-    hyprland.monitors.forEach((m) => (monitorMap[m.id] = m.name));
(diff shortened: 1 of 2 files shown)


orxngc:
@Jas-SinghFSU This broke the unoccupied workspaces option.
image

Jas-SinghFSU:
@orxngc This was intended. You can lower the `Total Workspaces` count to 0 if you don't want any inactive workspaces to show.

Jas-SinghFSU:
The goal of this PR was to allow a baseline # of workspaces to always show and the rest are dynamic.

orxngc:
@Jas-SinghFSU Ah okay, noted, thank you

PR #200, opened 2024-08-29
Nix hyprland settings exec-once to list of strings instead of single string

anotherhadi:
Using a list of strings for the exec-once field is "cleaner" and allows you to reference it in other files without overwriting the initial value of the exec-once field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,11 +145,9 @@ home.packages = with pkgs; [
 ];
 
 # or reference it directly in your Hyprland configuration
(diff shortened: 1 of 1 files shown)


anotherhadi:
Should also update the wiki for this one, don't know if the wiki is using the README.md or if it's a different file.

Jas-SinghFSU:
https://github.com/Jas-SinghFSU/hyprpanel-docs

Jas-SinghFSU:
As I don't use nix, is anyone that does able to review this PR?

Issue #152, opened 2024-08-18
Change number of shortcuts and directories


LDAP:
**Is your feature request related to a problem? Please describe.**
I find it overwhelming having too many shortcuts and directories.

**Describe the solution you'd like**
Adapt the layout to the number of items with non-empty configuration.

**Describe alternatives you've considered**
-

**Additional context**
![image](https://github.com/user-attachments/assets/5f342e1e-7584-4e8f-830f-a941bcdb89b0)

Jas-SinghFSU:
Shortcut icons now automatically hide if no command is provided.

PR #210, opened 2024-08-31
Shortcut buttons in the dashboard automatically hide and scale.

Jas-SinghFSU:
The shortcut buttons in the dialog menu automatically hide if a command is not provided in the settings dialog.


Jas-SinghFSU:
closes #152

PR #241, opened 2024-09-10
Improve Theme Filling Script

painerp:
- **Removed Obsolete Values**: Values that aren't in the complete theme anymore are now removed from all target themes.

- **Improved Log Readability**: The script now prints just the filename instead of the whole path.


Jas-SinghFSU:
Could there be a case that the 'complete' theme isn't actually fully complete and running the script would cause a regression in the other themes that are fully complete due to the property removal?

painerp:
This is indeed a possibility. If I assessed the script correctly it assumes the 'complete' theme (by default cappucino_mocha) to be the template and propagates all changes from it to the other themes.

In my opinion it isn't that big of a problem though since you clearly see the removed lines in the output of the script. We could potentially also change the color of the echo to be even more visible.

In the meantime, documenting this behavior would be a good idea, so that other contributors are aware of how the script works.

In the long run, automatically extracting all theme options would likely be wise, ensuring everything is present and making it less obscure which theme is the template.

PR #249, opened 2024-09-11
Add support for dnf and fedora in checkUpdates.sh

smathles:
Adds dnf package manager compatibility to checkUpdates.sh. In future I'm pondering adding support for flatpaks etc, but I think that's something that should first be discussed for implementation in arch/ubuntu/whatever else as well.

### DIFF
--- a/modules/bar/Bar.ts
+++ b/modules/bar/Bar.ts
@@ -279,6 +279,7 @@ export const Bar = (() => {
                 class_name: 'bar-panel-container',
                 child: Widget.CenterBox({
                     class_name: 'bar-panel',
+                    //css: 'padding: 1px 0px 0px 0px',
                     css: 'padding: 1px',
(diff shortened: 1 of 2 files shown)


modules/bar/Bar.ts
```diff
@@ -279,6 +279,7 @@ export const Bar = (() => {
                 class_name: 'bar-panel-container',
                 child: Widget.CenterBox({
                     class_name: 'bar-panel',
+                    //css: 'padding: 1px 0px 0px 0px',
```
Jas-SinghFSU, line 282:
```suggestion
```
I see this lil sucka got through lol

smathles: riiiiip yeah forgot about that

Jas-SinghFSU (approved):

PR #252, opened 2024-09-11
support "*" as wildcard layout key

j-brn:
allows specifing a default layout that applies for all monitors if no more specific layout is found.

The following layout configuration would apply the first layout to all monitors except "1" 

```json
{
  "*": {
    "left": [
      "workspaces"
    ],
    "middle": [
      "media"
    ],
    "right": [
      "clock",
      "notifications"
    ]
  },
  "1": {
    "left": [
      "dashboard",
      "workspaces",
      "windowtitle"
    ],
    "middle": [
      "media"
    ],
    "right": [
      "volume",
      "clock",
      "notifications"
    ]
  }
}
```

### DIFF
--- a/modules/bar/Bar.ts
+++ b/modules/bar/Bar.ts
@@ -65,10 +65,19 @@ type BarLayout = {
     [key: string]: Layout
 }
 
-const getModulesForMonitor = (monitor: number, curLayouts: BarLayout) => {
-    const foundMonitor = Object.keys(curLayouts).find(mon => mon === monitor.toString());
+const getLayoutForMonitor = (monitor: number, layouts: BarLayout): Layout => {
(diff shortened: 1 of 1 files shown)


j-brn:
new PR because github doesn't allow changing the source branch on pull requests

Jas-SinghFSU:
I wonder if we can make this an actual regex? Would that be useful?

j-brn:
I thought about this too, but the only real use case would be to do something like this 

```json
"(1|2|3)": $layout
```

IMHO regex is overkill here. If this is a feature we want, then I'd rather support comma/pipe separated lists as keys

Jas-SinghFSU:
Valid point. I'll test and merge once I get some time.

Jas-SinghFSU:
Thanks for the contribution :heart:

PR #253, opened 2024-09-11
calendar menu: make weather optional

j-brn:
Adds a setting to enable or disable the weather widget in the calendar menu.
Since the weather widget requires additional setup to function, the default is off.  

image
image


modules/menus/calendar/index.ts
```diff
@@ -19,7 +22,13 @@ export default () => {
             Widget.Box({
               class_name: "calendar-content-items",
               vertical: true,
-              children: [TimeWidget(), CalendarWidget(), WeatherWidget()],
+              children: weatherEnabled.bind("value").as(isWeatherEnabled => {
```
Jas-SinghFSU, line 25: While this works, I've discovered that you want to decouple binding as much as possible. In this case, the Time and Calendar are both coupled with that binding.

What we can do instead is move the binding into the WeatherWidget and render an empty box if the weather is disabled. That way the binding only effects the component it's meant to.

Everything else looks good though, you've got the right idea :+1:

Jas-SinghFSU (changes requested):

modules/menus/calendar/index.ts
j-brn: Personally, I wouldn't expect something like WeatherWidget to render an empty Box, because the clock menu is configured to not include it. 

My idea of the option was that it should be used to exclude it from the clock menu and not disable weather at all. The weather widget is independent and could also be used in a dedicated weather bar module.

j-brn: If the empty box is the wanted behavior, then the setting should be called differently

Jas-SinghFSU: By empty box I mean returning Widget.Box() which renders nothing at all. Visually it's the same effect as what you have, but if we want enable granularity in picking specific modules and reusing them then I think this approach is fine.

Jas-SinghFSU:
@j-brn There are merge conflicts in this file due to a huge linting and formatting PR.

I have gone ahead and created a branch that you should be able to merge into yours to resolve the PRs.

When you get a chance, feel free to merge with `rebase/253`.

j-brn:
> @j-brn There are merge conflicts in this file due to a huge linting and formatting PR.
> 
> I have gone ahead and created a branch that you should be able to merge into yours to resolve the conflicts.
> 
> When you get a chance, feel free to merge with `rebase/253`.

@Jas-SinghFSU rebased from master

Jas-SinghFSU:
@j-brn Could you merge with rebase/253 please? It includes some styling fixes like the extra padding below the calendar when weather is disabled.

j-brn:
> @j-brn Could you merge with rebase/253 please? It includes some styling fixes like the extra padding below the calendar when weather is disabled.

@Jas-SinghFSU done

Jas-SinghFSU (approved):

Jas-SinghFSU:
Looks good! Thanks for the contribution fam!

PR #254, opened 2024-09-14
add official arch linux repo updatecheck to script for updates customModules

TheyCallMeSecond:
the current checkUpdates.sh script just check AUR repo for update
i added official repo checkupdates to it

### DIFF
--- a/scripts/checkUpdates.sh
+++ b/scripts/checkUpdates.sh
@@ -1,24 +1,21 @@
 #!/bin/bash
 
 check_arch_updates() {
-    official_updates=$(checkupdates 2>/dev/null | wc -l)
-    aur_updates=$(yay -Qum 2>/dev/null | wc -l)
+    official_updates=0
+    aur_updates=0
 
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Could we make this a flag or a different command? :smile: 

Some users may only want to track pacman, aur, or both and I think it would be neat to be able to switch between the choices.

TheyCallMeSecond:
> Could we make this a flag or a different command? 😄
> 
> Some users may only want to track pacman, aur, or both and I think it would be neat to be able to switch between the choices.

separated now with flags

Jas-SinghFSU (changes requested):
Also, do we need to document that `pacman-contrib` needs to be installed to use checkupdates?

TheyCallMeSecond:
> Also, do we need to document that `pacman-contrib` needs to be installed to use checkupdates?

yes

Jas-SinghFSU (approved):

PR #262, opened 2024-09-15
adds power-profiles-daemon to dependencies

TheyCallMeSecond:
power profile switch in battery module needs power-profiles-daemon package


TheyCallMeSecond:
or is it not optional ?

Jas-SinghFSU:
I think having it as optional is fine but we should add a notification indicating that the user must install it to change profiles. I can do that in a separate pr.

PR #272, opened 2024-09-19
Fixes the window title label option

painerp:
- Fixes the label option so it actually hides the label when toggled off
- Adds a side effect to the label and icon toggles so one of them is always active


modules/bar/window_title/index.ts
```diff
@@ -2,6 +2,7 @@ const hyprland = await Service.import('hyprland');
 import { BarBoxChild } from 'lib/types/bar';
 import options from 'options';
 import { ActiveClient } from 'types/service/hyprland';
```
Jas-SinghFSU, line 4:
```suggestion
import { Child } from 'lib/types/widget';
import { ActiveClient } from 'types/service/hyprland';
```

modules/bar/window_title/index.ts
```diff
@@ -177,35 +178,29 @@ const ClientTitle = (): BarBoxChild => {
                     truncation_size.bind('value'),
                 ],
                 (client, useCustomTitle, useClassName, showLabel, showIcon, truncate, truncationSize) => {
+                    const children: Label<never>[] = [];
```
Jas-SinghFSU, line 181:
```suggestion
                    const children: Label<Child>[] = [];
```
Let's assign the right type to the Label.

Jas-SinghFSU (changes requested):
Other than than looks great! :+1:

Jas-SinghFSU:
:heart:

Jas-SinghFSU (approved):

Issue #298, opened 2024-10-03
CLI: Get status of window


anotherhadi:
**Is your feature request related to a problem? Please describe.**
I'm creating a script that disable border rounding, gaps, etc to focus on my work and I want to hide my bar.
I currently just pkill ags but that's not very clean, and I can't use the "--toggle-window bar-0" because I don't know the status.

**Describe the solution you'd like**
A quick command that take the window name (bar-0, dashboard, ...) and return 0/1, true/false, ...
`hyprpanel -s (--status) [window]`

**Describe alternatives you've considered**
pgrep ags

Jas-SinghFSU:
Added a cli command to check window status. Please refer to the description of the MR until I update the documentation for instructions.
https://github.com/Jas-SinghFSU/HyprPanel/pull/301

PR #301, opened 2024-10-05
Added a CLI method to check the visibility of a window.

Jas-SinghFSU:
For example, a bar can be checked if it's visible or now via:
```bash
ags -r "isWindowVisible('bar-0')"
```
For bars it:
```
bar-n
```
Where `n` is the index of your monitor (0 is your primary one usually, so `bar-0`).

For menus such as the volume, dashboard or media menu, you can refer to the menu names here https://hyprpanel.com/configuration/cli.html#toggling-menus.

For example, to check if the dashboard menu is active you can type:

```bash
ags -r "isWindowVisible('dashboardmenu')"
```


Jas-SinghFSU:
closes #298

PR #315, opened 2024-10-08
fix: nix flakes install

### LINKED ISSUE #291: Fail to build on NixOS due to gnome-bluetooth
When i try to build, i get the following error:
`evaluation aborted with the following error message: 'lib.customisation.callPackageWith: Function called without required argument "gnome-bluetooth" at /nix/store/3p6803sfb55m8xcdcsx2alayfqf6nphs-source/nix/default.nix:24'`

M0NsTeRRR:
Fix #291 (as discussed on discord)


Jas-SinghFSU:
I would be extra aware that the `gnome.` prefix is set to be removed soon. From what I gather it's already removed on unstable and should be the moving to stable in the future.

Perhaps we update the documentation to reflect the installation for the current stable nix environment that way we don't break it for people using unstable.

zDyanTB:
```nix
(com.github.Aylur.ags:7922): Gjs-Console-CRITICAL **: 21:09:42.768: Error running command "$HOME/.config/ags/scripts/checkUpdates.sh -arch": bash: line 1: /home/zdyant/.config/ags/scripts/checkUpdates.sh: No such file or directory

(com.github.Aylur.ags:7922): Gjs-Console-CRITICAL **: 21:09:42.771: net.hadess.PowerProfiles is not available

(com.github.Aylur.ags:7922): GLib-GObject-CRITICAL **: 21:09:43.185: value "nan" of type 'gdouble' is invalid or out of range for property 'value' of type 'gdouble'

(com.github.Aylur.ags:7922): Gjs-WARNING **: 21:09:43.208: Unhandled promise rejection. To suppress this warning, add an error handler to your promise chain with .catch() or a try-catch block around your await expression. Stack trace of the failed promise:
execAsync@resource:///com/github/Aylur/ags/utils/exec.js:78:12
onConfigParsed@file:///nix/store/j2jmkvc01njmfavfxjbwawqy27nvg12w-hyprpanel-latest/config.js:15238:31
_load@resource:///com/github/Aylur/ags/app.js:203:29
async*vfunc_activate@resource:///com/github/Aylur/ags/app.js:109:14
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34


(com.github.Aylur.ags:7922): GLib-GObject-CRITICAL **: 21:09:44.358: value "nan" of type 'gdouble' is invalid or out of range for property 'value' of type 'gdouble'
Gdk-Message: 21:09:44.495: Lost connection to Wayland compositor.
```

kashw2:
Why don't we just update the nixpkgs input to follow master instead of solutions that break this package for one portion of the user base to fix it for another?

This would remove ambiguity and would put the line in the sand saying that "This package follows whatever the latest is on nixpkgs master as of the time of the last flake update, you are free to change `inputs.nixpkgs.follows` as you please however your mileage may vary" as opposed to "This package follows nixos-unstable which is not latest and can cause issues for anyone who specifies their inputs as anything else". I believe that any project like this strives to be up to date with it's dependencies and upstream changes should follow master for those latest changes and set a firm line in the sand on that. Doing so would remove the issue from this project and return it to the user and the solution would no longer be to make pull requests that revert the previous fixes made for one half of the user base to appease the other.

Jas-SinghFSU:
@kashw2 I'm aligned with that approach. I don't particularly enjoy playing hot potato between unstable and stable either so anything to update our nix config to bring it more inline with best practices, I'm on board with.

Would you be able to make a PR?

M0NsTeRRR:
> Why don't we just update the nixpkgs input to follow master instead of solutions that break this package for one portion of the user base to fix it for another?

I don't agree with you. Some packages manage to handle multiple NixOS channels (though I can't explain exactly how, as I'm still a beginner). We shouldn't pin to `master` but to the `stable` or `unstable` branch, and I would recommend `stable` if we have to choose. Pinning to `master` will only work for `unstable`, since in `stable` we don't have `gnome.gnome-bluetooth` yet.

kashw2:
> > Why don't we just update the nixpkgs input to follow master instead of solutions that break this package for one portion of the user base to fix it for another?
> 
> I don't agree fully with you. Some packages manage to handle multiple NixOS channels (though I can't explain exactly how, as I'm still a beginner). We shouldn't pin to `master` see [here](https://nixos.wiki/wiki/Nix_channels) but to the `stable` or `unstable` branch, and I would recommend `stable` if we have to choose. Pinning to `master` will only work for `unstable`, since in `stable` we don't have `gnome.gnome-bluetooth` yet.
> 
> We can find a way that works for the two channels.
> 
Either channel works however the issue at hand isn't going to be fixed by changing between gnome.gnome-bluetooth and gnome-bluetooth constantly. The fix is pinning to a channel and making the user base aware that they may have issues if they alter it inside of their flake.

The decision over what channel to follow is ultimately at the discretion of @Jas-SinghFSU. The nature of flakes gives us the flexibility to use whatever channel the maintainers of this project want without any issue due to pinning in the lockfile assuming that whenever it is updated it is adequately tested. I suggest master because it allows us to be as up to date as we can be with changes coming from upstream nixpkgs.

M0NsTeRRR:
> Either channel works however the issue at hand isn't going to be fixed by changing between gnome.gnome-bluetooth and gnome-bluetooth constantly.
I fully agree with you. This PR is more of a test, as we discussed on Discord while trying to figure out why it wasn't working.

>  I suggest master because it allows us to be as up to date as we can be with changes coming from upstream nixpkgs.

Master is not a Nix channel and doesn't benefit from unstable CI tests. As stated, *'Contrary to a user of the git master branch, a channel user will benefit from both verified commits and binary packages from the binary cache.'* Therefore, master should not be considered a reliable source. 

Reference: [Nix Channels](https://nixos.wiki/wiki/Nix_channels), [Nixpkg documentation](https://nixos.org/manual/nixpkgs/stable/#overview-of-nixpkgs).

Of course, it's up to the maintainer to decide between `stable` or `unstable`.

> From my testing, this is a bit of PEBKAC where users are trying to pin this flakes input to their own nixpkgs input which follows a different channel that doesn't have changes to the top level attributes applied yet. I have been able to get this package to build using a version pinned to a version of this packages flake that used gnome.gnome-bluetooth and one that used just gnome-bluetooth. The only difference was what I made my nixpkgs input follow and making sure I invalidated the previous build and its store path before rebuilding.

`follows` is a good practice as it helps avoid having multiple nixpkgs on your machine. From my point of view, this is not a PEBKAC issue. The documentation doesn't mention anything about pinning inputs, so we can't expect new users to know this from the start.

Also, as I mentioned, how do other packages with this issue handle it? I'm pretty sure this isn't uncommon, and some flakes manage it within the flake itself to support both stable and unstable versions.

kashw2:
> > Either channel works however the issue at hand isn't going to be fixed by changing between gnome.gnome-bluetooth and gnome-bluetooth constantly.
> 
> I fully agree with you. This PR is more of a test, as we discussed on Discord while trying to figure out why it wasn't working.
> 
> >  I suggest master because it allows us to be as up to date as we can be with changes coming from upstream nixpkgs.
> 
> Master is not a Nix channel and doesn't benefit from unstable CI tests. As also stated, *'Contrary to a user of the git master branch, a channel user will benefit from both verified commits and binary packages from the binary cache.'* Therefore, master should not be considered a reliable source. 
> 
> Reference: [Nix Channels](https://nixos.wiki/wiki/Nix_channels), [Nixpkg documentation](https://nixos.org/manual/nixpkgs/stable/#overview-of-nixpkgs).
> 
> Of course, it's up to the maintainer to decide between `stable` or `unstable`.
> 
> > From my testing, this is a bit of PEBKAC where users are trying to pin this flakes input to their own nixpkgs input which follows a different channel that doesn't have changes to the top level attributes applied yet. I have been able to get this package to build using a version pinned to a version of this packages flake that used gnome.gnome-bluetooth and one that used just gnome-bluetooth. The only difference was what I made my nixpkgs input follow and making sure I invalidated the previous build and its store path before rebuilding.
> 
> `follows` is a good practice as it helps avoid having multiple nixpkgs on your machine. From my point of view, this is not a PEBKAC issue. The documentation doesn't mention anything about pinning inputs, so we can't expect new users to know this from the start.
> 
> Also, as I mentioned, how do other packages with this issue handle it? I'm pretty sure this isn't uncommon, and some flakes manage it within the flake itself to support both stable and unstable versions.

The more I thought about it, after the last comment the more I tend to agree that unstable is the best place for the flake to follow.

This issue specifically I would say is common and somewhat to be expected with the way the projects flake is currently set up in situations where nixpkgs makes changes to top-level attributes.

On the topic of letting users know, this is what could change and what I would champion for to avoid this. Imo it should go without saying that one of the first points of call to try and fix this should be to use the nixpkgs input that this project uses instead of pointing it to their own. Although I understand why some would like to keep all of their inputs pinned to a common input to save space. Although I wouldn't necessarily call this a good or bad practice as it gives the impression that not doing it is bad when it's apart of the problem in this situation.

I'm on board with dual channel solution although I haven't seen many flakes out there that have dual channel inputs.

Issue #314, opened 2024-10-08
Hide the pagination on the notification menu when page == 1


anotherhadi:
Everything is in the title!

rubiin:
I think the correct feature would be to hide the pager when there are only one page .

PR #335, opened 2024-10-17
fix: hide pager when only first page

rubiin:
Hide the page when there are only one page

when 0 notifications == no pager
image

when has only first page === no pager
image

when there is more than one pages === pager
image

closes https://github.com/Jas-SinghFSU/HyprPanel/issues/314


Jas-SinghFSU:
Thanks :tada:

PR #338, opened 2024-10-18
feature: hide notifications count when 0

rubiin:


Jas-SinghFSU:
I like this but we should make it a toggleable option.

rubiin:
@Jas-SinghFSU sure thing

rubiin:
could you check again

rubiin:
@Jas-SinghFSU could you check again

widget/settings/pages/config/bar/index.ts
```diff
@@ -738,6 +738,11 @@ export const BarSettings = (): Scrollable<Gtk.Widget, Gtk.Widget> => {
                     title: 'Show Total # of notifications',
                     type: 'boolean',
                 }),
+                Option({
+                    opt: options.bar.notifications.hideCountWhenZero,
+                    title: 'Hide Total # of notifications when zero',
```
Jas-SinghFSU, line 743:
```suggestion
                    title: 'Auto Hide Label',
                    subtitle: 'Hide Total # of notifications when zero',
```

Jas-SinghFSU:
Good stuff! The only issue is that if you are using split style (pick a split theme for instance) and hide the label, it has sharp edges and isn't accounting for the radius change.

We can remedy this by adding
```ts
        &:last-child {
            border-radius: $bar-buttons-radius;
        }
```
in our `scss/style/bar/notifications.scss` within the `.style2` scope. Ex:
```ts
    .bar-button-icon.notifications {
        border-top-left-radius: $bar-buttons-radius;
        border-bottom-left-radius: $bar-buttons-radius;
        background: if($bar-buttons-monochrome, $bar-buttons-icon, $bar-buttons-notifications-icon_background);
        padding: $bar-buttons-padding_y 0em;
        padding-left: $bar-buttons-padding_x;
        padding-right: $bar-buttons-notifications-spacing;
        border-top-left-radius: if(
            $bar-buttons-notifications-enableBorder or $bar-buttons-enableBorders,
            $bar-buttons-radius * 0.4,
            $bar-buttons-radius
        );
        border-bottom-left-radius: if(
            $bar-buttons-notifications-enableBorder or $bar-buttons-enableBorders,
            $bar-buttons-radius * 0.4,
            $bar-buttons-radius
        );
        color: if($bar-buttons-monochrome, $bar-buttons-icon, $bar-buttons-notifications-icon);

        &:last-child {
            border-radius: $bar-buttons-radius;
        }
    }
```

modules/bar/notifications/index.ts
```diff
@@ -49,6 +55,9 @@ export const Notifications = (): BarBoxChild => {
                         });
 
                         if (showTotal) {
+                            if (hideCountForZero && notif.length === 0) {
```
Jas-SinghFSU, line 58:
```suggestion
                            if (hideCountForZero && filteredNotifications.length === 0) {
```
Also overlooked that we support notifications filters. So we want to check based on the total count reflective of our filtered notifications. Otherwise you might have some notifications blocked by the filter and see 0 notifications, but the label isn't hidden since we're checking the unfiltered count.

Other than that this is ready to merge :)

PR #342, opened 2024-10-20
feature: remove delay when clearing all notifications

rubiin:
When having 50+ notifications, clicking the clear all button takes some time before it does the job.

This was because we were adding a delay with settimeout of 100 ms . 
This makes the clearall instant which is a better UX similar to other implementations like in `swaync`

### DIFF
--- a/widget/settings/pages/config/notifications/index.ts
+++ b/widget/settings/pages/config/notifications/index.ts
@@ -48,7 +48,9 @@ export const NotificationSettings = (): Scrollable<Child, Attribute> => {
                 Option({
                     opt: options.notifications.clearDelay,
                     title: 'Clear Delay',
-                    subtitle: 'The delay in milliseconds before a notification is cleared',
+                    subtitle:
+                        'The delay in milliseconds before a notification is cleared' +
+                        '\nWARNING: Setting this value too low may crash AGS depending on your system.',
                     type: 'number',
                     increment: 20,
 
--- a/widget/settings/pages/config/notifications/index.ts
+++ b/widget/settings/pages/config/notifications/index.ts
@@ -53,7 +53,6 @@ export const NotificationSettings = (): Scrollable<Child, Attribute> => {
                         '\nWARNING: Setting this value too low may crash AGS depending on your system.',
                     type: 'number',
                     increment: 20,
-
                 }),
                 Option({
                     opt: options.notifications.timeout,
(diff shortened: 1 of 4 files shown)


Jas-SinghFSU:
I'm not a fan of the delay either ideally it should be MUCH lower but this is something you have to do to prevent crashes in AGS. Notice the `clearDelay` property on the AGS documentation https://aylur.github.io/ags-docs/services/notifications/.

For some users this number has to be even higher. If we want to remove it, we can add an option to let the user define the delay amount and add a warning below the settings that "Setting this value too low may crash AGS".

rubiin:
Looks like this PR still needs some work. I will try to add in the option by tomorrow

rubiin:
Updated the setting widget to include the delay with the initial value as `100 ms`.

widget/settings/pages/config/notifications/index.ts
```diff
@@ -45,6 +45,12 @@ export const NotificationSettings = (): Scrollable<Child, Attribute> => {
                     subtitle: 'The notification will follow the monitor of your cursor',
                     type: 'boolean',
                 }),
+                Option({
+                    opt: options.notifications.clearDelay,
+                    title: 'Clear Delay',
+                    subtitle: 'The delay in milliseconds before a notification is cleared',
+                    type: 'number',
```
Jas-SinghFSU, line 52:
```suggestion
                    type: 'number',
                    increment: 20,

```
We can add a default increment size of 20 so the button is easier to configure (20ms per click).

widget/settings/pages/config/notifications/index.ts
```diff
@@ -45,6 +45,12 @@ export const NotificationSettings = (): Scrollable<Child, Attribute> => {
                     subtitle: 'The notification will follow the monitor of your cursor',
                     type: 'boolean',
                 }),
+                Option({
+                    opt: options.notifications.clearDelay,
+                    title: 'Clear Delay',
+                    subtitle: 'The delay in milliseconds before a notification is cleared',
```
Jas-SinghFSU, line 51:
```suggestion
                    subtitle:
                        'The delay in milliseconds before a notification is cleared' +
                        '\nWARNING: Setting this value too low may crash AGS depending on your system.',
```

Jas-SinghFSU:
That's for the wonderful contributions! :heart:

PR #343, opened 2024-10-20
feat: add option to hide seconds on calendar

rubiin:
Allow option to hide seconds on calendar, backward compatible as false is the default value

image

image


Jas-SinghFSU:
LGTM. Thank you for the contribution :heart:!!!

Issue #165, opened 2024-08-22
Bluetooth battery


shadowe1ite:
is there anyways to enable bluetooth battery percentage next to icon (left, righ)

Jas-SinghFSU:
No, this would be a feature request.

PR #352, opened 2024-10-21
feat: show battery on bluetooth device list

rubiin:
closes https://github.com/Jas-SinghFSU/HyprPanel/issues/165
image

For connected devices only


smathles:
Very keen for this. Would be awesome to monitor device battery levels instead of just the system!

Jas-SinghFSU:
How do we handle scenarios where battery is not available? I believe not all devices provide a battery status.

rubiin:
Let me handle the scenerio too

Jas-SinghFSU:
Is null a response that we get back? I have a bluetooth device which doesn't have battery levels but it's now showing null... Rather it shows `battery-level: 0` when no battery is available or `battery-level: 1` when a battery can be fetched. Could you verify with devices if you get the same thing between ones with battery status and those without?

rubiin:
I dont have a device without a battery so cannot test this properly.

Jas-SinghFSU:
I would like it to have a toggle option.

Additionally, I think we should use a `Widget.Separator` and set the spacing between the status and the battery separator via margins. I've used spaces in the past and based on the font you're using it can be very inconsistent.

Jas-SinghFSU:
Let's also separate the battery icon and the percentage into their own Widget.Labels. Reason is that we can then assign a `txt-icon` class to the battery icon (`txt-icon` is built to properly align nerdfont icons) to center/align it properly.

rubiin:
@Jas-SinghFSU Added a dedicated settings menu for bluetooth module. From there user can toggle the bluetooth battery as well as configure the icon to use

rubiin:
> I dont have a device without a battery so cannot test this properly. @Jas-SinghFSU if you have a device perhaps you can test and tell me the op. I will adjust the code accordingly
> 
> Also do we want it to have a toggle option?

I will leave the testing to you for device without battery as I cannot test it out due to this

Jas-SinghFSU:
Sounds good, thank you! I'll try to get to it tomorrow.

smathles:
image
All my devices have the battery percentage showable, else I'd test it :(. Might just be up to you sorry @Jas-SinghFSU.

modules/menus/bluetooth/devices/devicelist.ts
```diff
@@ -101,10 +104,53 @@ const devices = (bluetooth: Bluetooth, self: Box<Gtk.Widget, unknown>): Box<Chil
...
+                                                                                    : '',
+                                                                        }),
+                                                                        Widget.Label({
+                                                                            class_name: 'connection-status battery',
+                                                                            label:
+                                                                                device.connected &&
+                                                                                device.battery_percentage !== null
+                                                                                    ? `${device.battery_percentage}%`
+                                                                                    : '',
+                                                                        }),
```
Jas-SinghFSU, line 150: I think a safe bet here then if we don't know for certain if battery percentage exists is to add a safe conditional like
```ts
!isNaN(device.battery_percentage) && device.battery_percentage > 0 
    ? `${device.battery_percentage}%`
    : '',
```

Jas-SinghFSU: We can add this everywhere we rely on the battery to be shown. Also we need to make sure that we don't generate the divider if no battery is present (using the above conditional).

rubiin: lets make a function for this check. 

````
typeof device.battery_percentage === 'number' && device.battery_percentage >= 0 
    ? `${device.battery_percentage}%`
    : '',
```

And add to divider, battery icon and battery percent label

rubiin: is there a way we could perhaps group the divider, battery icon and battery percent label. Then we can apply this condition to group rather than on individual items as we are doing now.

Jas-SinghFSU: We can just put the entire battery related widget set in a box and make the box generate empty children if the battery for that item doesn't exist. Ex:
```ts
Widget.Box({
children: if-battery-exists ? [render the separator, label, label] : []
})
```

rubiin: will try this. This will make the code a lot cleaner

options.ts
```diff
@@ -1054,6 +1054,9 @@ const options = mkOptions(OPTIONS, {
     menus: {
         transition: opt<Transition>('crossfade'),
         transitionTime: opt(200),
+        bluetooth: {
+            showBattery: opt(false),
+            batteryIcon: opt('󰥉'),
         volume: {
```
Jas-SinghFSU, line 1060:
```suggestion
        bluetooth: {
            showBattery: opt(false),
            batteryIcon: opt('󰥉'),
        },
        volume: {
```

modules/menus/bluetooth/devices/devicelist.ts
```diff
@@ -101,10 +104,57 @@ const devices = (bluetooth: Bluetooth, self: Box<Gtk.Widget, unknown>): Box<Chil
...
+                                                                                              class_name:
+                                                                                                  'connection-status txt-icon',
+                                                                                              label: `${batIcon}`,
+                                                                                          }),
+                                                                                          Widget.Label({
+                                                                                              class_name:
+                                                                                                  'connection-status battery',
+                                                                                              label: `${device.battery_percentage}%`,
+                                                                                          }),
+                                                                                      ],
```
Jas-SinghFSU, line 153:
```suggestion
                                                                                device.battery_percentage >= 0
                                                                                    ? [
                                                                                          Widget.Separator({
                                                                                              class_name:
                                                                                                  'menu-separator',
                                                                                          }),
                                                                                          Widget.Label({
                                                                                              class_name:
                                                                                                  'connection-status txt-icon',
                                                                                              label: `${batIcon}`,
                                                                                          }),
                                                                                          Widget.Label({
                                                                                              class_name:
                                                                                                  'connection-status battery',
                                                                                              label: `${device.battery_percentage}%`,
                                                                                          }),
                                                                                      ],
                                                                                    : []
```

modules/menus/bluetooth/devices/devicelist.ts
```diff
@@ -101,10 +104,57 @@ const devices = (bluetooth: Bluetooth, self: Box<Gtk.Widget, unknown>): Box<Chil
...
```
Jas-SinghFSU, line 152:
```suggestion
                                                                                      ]
```

PR #354, opened 2024-10-22
feat: add some info about connected wifi

rubiin:
Adds some stats similar to waybar on wifi

image

Totally opt in
image


Jas-SinghFSU:
There are scenarios where the AGS network service returns:
1. Name as an empty string `""`
2. Strength as `-1` indicating the lack of a valid strength (think null)
3. Frequency as `-1` indicating the lack of a valid frequency

We should handle all these scenarios where if these values are captured, we display `--` instead. So if no wifi is connected:
```
Network: None
Signal Strength: --
Frequency: --
```
Or something similar. Essentially we should handle invalid values.

Jas-SinghFSU:
We still need to handle the above scenarios. In case that you're not connected to a Wifi, it will show a negative value for the signal and frequency. We can instead format it like this when invalid values are present:
```
Network: None
Signal Strength: --
Frequency: --
```

rubiin:
sure thing , will update that asap

rubiin:
@Jas-SinghFSU added fallback

modules/menus/network/wifi/WirelessAPs.ts
```diff
@@ -178,35 +178,77 @@ const renderWAPs = (
...
-                                                .catch((err) => {
-                                                    connecting.value = '';
-                                                    console.error(`Error while forgetting "${ap.ssid}": ${err}`);
+                            child: Widget.Box({
+                                children: [
+                                    Widget.Button({
+                                        class_name: 'menu-icon-button network',
+                                        child: Widget.Label({
+                                            tooltip_text: 'Disconnect',
+                                            class_name: 'menu-icon-button network disconnect',
```
Jas-SinghFSU, line 187:
```suggestion
                                            class_name: 'menu-icon-button network disconnect txt-icon',
```
We always have to assign txt-icon class to Label based nerdfont icons so they align properly :smile:

modules/menus/network/wifi/WirelessAPs.ts
```diff
@@ -178,35 +178,77 @@ const renderWAPs = (
...
-
-                                            Utils.execAsync(`nmcli connection delete ${connectionId} "${ap.ssid}"`)
-                                                .then(() => (connecting.value = ''))
-                                                .catch((err) => {
-                                                    connecting.value = '';
-                                                    console.error(`Error while forgetting "${ap.ssid}": ${err}`);
+                            child: Widget.Box({
+                                children: [
+                                    Widget.Button({
+                                        class_name: 'menu-icon-button network',
```
Jas-SinghFSU, line 184:
```suggestion
                                        class_name: 'menu-icon-button network disconnect',
```

modules/menus/network/wifi/WirelessAPs.ts
rubiin: :D I forgot to add that in a rush

modules/bar/network/index.ts
```diff
@@ -68,7 +79,8 @@ const Network = (): BarBoxChild => {
                             }
                             return Widget.Label({
                                 class_name: 'bar-button-label network-label',
-                                label: wfi.ssid ? `${trunc ? wfi.ssid.substring(0, tSize) : wfi.ssid}` : '--',
+                                label: wfi.ssid ? `${trunc ? wfi.ssid.substring(0, tSize) : wfi.ssid}` : '',
```
Jas-SinghFSU, line 82:
```suggestion
                                label: wfi.ssid ? `${trunc ? wfi.ssid.substring(0, tSize) : wfi.ssid}` : '--',
```

Jas-SinghFSU (approved):

Jas-SinghFSU:
Awesome stuff :heart:

PR #356, opened 2024-10-23
fix: make bottom padding consistent for dashboard

rubiin:
when disabling the sections  on dashboard , the bottom padding is also gone

image

this makes the padding consistent across the disable operations of sections in dashboard

image


Jas-SinghFSU:
We shouldn't add static padding this way because if the element is visible, it will double the padding. Ex
image

I will handle this in a different ticket. Basically we have to count the children in css and do:
1. If first child then add larger padding to top
2. If last child then add larger padding to bottom
3. Otherwise add normal bottom padding

rubiin:
true. i will close this in favor

PR #361, opened 2024-10-23
fix: dont hardcode file manager and make use of $TERM

rubiin:
With this, user can use any file manager they have setup.
This also removes terminal and uses the one setup on $TERM varibale. The original PR contains a mistake as TERM env is used instead of TERMINAL

https://www.ibm.com/docs/en/informix-servers/14.10?topic=products-term-environment-variable-unix
https://bash.cyberciti.biz/guide/$TERM_variable

closes https://github.com/Jas-SinghFSU/HyprPanel/pull/156

### DIFF
--- a/modules/menus/dashboard/stats/index.ts
+++ b/modules/menus/dashboard/stats/index.ts
@@ -24,7 +24,7 @@ interval.connect('changed', () => {
 
 const handleClick = (): void => {
     App.closeWindow('dashboardmenu');
-    Utils.execAsync(`bash -c "$TERMINAL -e btop"`).catch((err) => `Failed to open btop: ${err}`);
+    Utils.execAsync(`bash -c "${terminal} -e btop"`).catch((err) => `Failed to open btop: ${err}`);
 };
 
 const Stats = (): BoxWidget => {
--- a/modules/menus/dashboard/stats/index.ts
+++ b/modules/menus/dashboard/stats/index.ts
@@ -7,6 +7,7 @@ import Cpu from 'services/Cpu';
 import Storage from 'services/Storage';
 import { renderResourceLabel } from 'customModules/utils';
 
+const { terminal } = options;
 const { enable_gpu, interval } = options.menus.dashboard.stats;
 
 const ramService = new Ram();
--- a/modules/menus/dashboard/stats/index.ts
+++ b/modules/menus/dashboard/stats/index.ts
@@ -7,7 +7,6 @@ import Cpu from 'services/Cpu';
 import Storage from 'services/Storage';
 import { renderResourceLabel } from 'customModules/utils';
 
-const { terminal } = options;
 const { enable_gpu, interval } = options.menus.dashboard.stats;
 
 const ramService = new Ram();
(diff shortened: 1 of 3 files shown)


rubiin:
@Jas-SinghFSU check this . I think this can be merged

options.ts
```diff
@@ -1160,8 +1160,6 @@ const options = mkOptions(OPTIONS, {
 
     scalingPriority: opt<ScalingPriority>('gdk'),
 
-    terminal: opt('kitty'),
-
     tear: opt(false),
```
Jas-SinghFSU, line 1163:
```suggestion
    terminal: opt('$TERMINAL'),
    
    tear: opt(false),
```

modules/menus/dashboard/stats/index.ts
```diff
@@ -7,7 +7,6 @@ import Cpu from 'services/Cpu';
 import Storage from 'services/Storage';
 import { renderResourceLabel } from 'customModules/utils';
 
-const { terminal } = options;
 const { enable_gpu, interval } = options.menus.dashboard.stats;
```
Jas-SinghFSU, line 10:
```suggestion
const { terminal } = options;
const { enable_gpu, interval } = options.menus.dashboard.stats;
```

modules/menus/dashboard/stats/index.ts
```diff
@@ -23,6 +22,11 @@ interval.connect('changed', () => {
     storageService.updateTimer(interval.value);
 });
 
+const handleClick = (): void => {
+    App.closeWindow('dashboardmenu');
+    Utils.execAsync(`bash -c "$TERMINAL -e btop"`).catch((err) => `Failed to open btop: ${err}`);
```
Jas-SinghFSU, line 27:
```suggestion
    Utils.execAsync(`bash -c "${terminal} -e btop"`).catch((err) => `Failed to open btop: ${err}`);
```

Jas-SinghFSU: We can leave these options in but have the default terminal configured in the options to '$TERMINAL'. That way if the user doesn't have a default terminal set, and the click doing nothing, they can instead configure it here to change that.

widget/settings/pages/config/general/index.ts
```diff
@@ -34,12 +34,6 @@ export const BarGeneral = (): Scrollable<Child, Attribute> => {
                         themeOnly: false,
                     },
                 }),
-                Option({
-                    opt: options.terminal,
-                    title: 'Terminal',
-                    subtitle: "Tools such as 'btop' will open in this terminal",
-                    type: 'string',
-                }),
                 Option({
```
Jas-SinghFSU, line 37:
```suggestion
                Option({
                    opt: options.terminal,
                    title: 'Terminal',
                    subtitle: "Tools such as 'btop' will open in this terminal",
                    type: 'string',
                }),
                Option({
```

rubiin:
@Jas-SinghFSU  fixed

Jas-SinghFSU (approved):

Issue #296, opened 2024-10-02
[FEATURE] Is it possible to window rewrite workspace icons?


j5onrf:
**Describe the solution you'd like**
When you visit a website the icon changes dynamic.

```
Waybar example: using "title"

"window-rewrite": {
		"title<.*youtube.*>": "",
                "class<firefox>": ""
}
```

**Additional context**
```
{
  "1": { "icon": "󰄛", "title<.*youtube.*>": ""},
}

{"1":"title<.*youtube.*>": "","2":"󰝰","3":"","4":"","5":""}
```
```
Window Titles
[
  ["youtube", "󰄛", "Youtube"],
]
```

Jas-SinghFSU:
You want to show the workspace icon based on the last active window in the workspace (defined as a regex)? Or are you talking about the window title? Or both lol.

I can see this being useful, it's just that we should probably split this into two feature requests if that's the case.

Jas-SinghFSU:
implemented by #368

PR #368, opened 2024-10-24
feat(workspace): map client classes to application icons

albohlabs:
I wanted to be able to dynamically map icons to Hyperland client classes. So I hacked together this solution for the waybar -> my own ags config and now HyprPanel. 

To test the functionality you need to enable the `Map Workspaces to Application Icons` option in the settings and paste the following json object in the `Application Icon Mappings`:

```json
{
  "code": "󰨞",
  "dbeaver": "",
  "foot": "",
  "evince": "",
  "footclient": "",
  "kitty": "",
  "wezterm": "",
  "org.wezfurlong.wezterm": "",
  "firefox": "",
  "firefox-developer-edition": "",
  "google-chrome": "",
  "chromium": "",
  "edge": "󰇩",
  "telegram": "",
  "thunderbird": "",
  "thunderbird-beta": "",
  "slack": "",
  "keymapp": "",
  "1Password": "",
  "obsidian": "󰠮",
  "keepassxc": "",
  "rofi": "",
  "Spotify": "󰓇",
  "spotube": "󰓇",
  "steam": "",
  "vlc": "󰕼",
  "draw.io": "󰇟",
  "org.qbittorrent.qBittorrent": ""
}
```

Pls give me some feedback if this kind of change is welcome and how i could improve to get the code merged.


rubiin:
any reason `org.qbittorrent.qBittorrent": ""` is a class while other seem to be normal string?

Jas-SinghFSU:
This is phenomenal! I do have some things I think we can improve on so I'll look at this more in depth this Friday/Weekend and leave comments here.

albohlabs:
@rubiin 

Not rly. When i run `hyprctl clients -j | jq -r ".[].initialClass"` to get all the classes i can check against, there is no common format in which they are written:

```bash
> hyprctl clients -j | jq -r ".[].initialClass"
firefox
org.telegram.desktop
firefox-developer-edition
thunderbird-beta
org.wezfurlong.wezterm
```

Jas-SinghFSU:
@albohlabs A couple of ways we could make it better:
1. Move the logic to get an icon into a function in `utils.ts`
2. Provide a setting to deduplicate windows
    1. This setting shows the app icon only once instead of one for each instance
3. Provide a setting to define the empty workspace icon
4. Provide separate/distinct parameters for these settings. Ex:
```ts
                                label: renderLabel(
                                    showIcons,
                                    available,
                                    active,
                                    occupied,
                                    showApplicationIcons,
                                    appIcons,
                                    workspaceMask,
                                    showWsIcons,
                                    wsIconMap,
                                    i,
                                    index,
                                    monitor,
                                    monitors,
                                ),
```
5. Handle these separately in the `renderLabel` function. Ex:
```ts
export const renderLabel = (
    showIcons: boolean,
    available: string,
    active: string,
    occupied: string,
    showAppIcons: boolean,
    appIcons: string | undefined,
    workspaceMask: boolean,
    showWsIcons: boolean,
    wsIconMap: WorkspaceIconMap,
    i: number,
    index: number,
    monitor: number,
    monitors: Monitor[],
): string => {
    if (showIcons) {
        if (hyprland.active.workspace.id === i || isWorkspaceActiveOnMonitor(monitor, monitors, i)) {
            return active;
        }
        if ((hyprland.getWorkspace(i)?.windows || 0) > 0) {
            return occupied;
        }
        if (monitor !== -1) {
            return available;
        }
    }

    if (showAppIcons) {
        return appIcons || '';
    }
    if (showWsIcons) {
        return getWsIcon(wsIconMap, i);
    }
    return workspaceMask ? `${index + 1}` : `${i}`;
};
```
6. Provide a way to define regular expressions for classnames in the map
7. Provide a default set of window mappings in code that can be overridden by custom ones

Let me know if that makes sense and if you want to take it on. Otherwise I can help as well.

albohlabs:
No problem. I can take it on. 

I just have one suggestion.

6. With respect to https://github.com/Jas-SinghFSU/HyprPanel/issues/296 it would make sense to wrap the regex in the property that we want to match again:
```json
{
  "org.wezfurlong.wezterm": ""
  "[dD]iscord": "󰙯",
  "class<*.telegram.*>": "",
  "title<.*youtube.*>": "",
}
```

Jas-SinghFSU:
@albohlabs I think that's the best approach. We can probably do what hyprland does and define it as:
```
"class:yourRegexHere": "",
"title:yourRegexHere": ""
```

albohlabs:
I've updated the PR. Pls have a look.

rubiin (approved):

modules/bar/workspaces/utils.ts
```diff
@@ -68,6 +68,70 @@ export const getWsColor = (
...
+    removeDuplicateIcons: boolean,
+    {
+        iconMap,
+        defaultIcon,
+        emptyIcon,
+    }: {
+        iconMap: ApplicationIcons;
+        defaultIcon: string;
+        emptyIcon: string;
+    },
```
Jas-SinghFSU, line 82: Nitpick: Prefer to keep type definitions in a type file.

Jas-SinghFSU, line 82: Nitpick: Prefer to keep type definitions in a type file.

modules/bar/workspaces/utils.ts
```diff
@@ -68,6 +68,70 @@ export const getWsColor = (
...
+) => {
+    // detect the clients attributes on the current workspace
+    const clients: ReadonlyArray<[className: string, title: string]> = hyprland.clients
```
Jas-SinghFSU, line 85: Nitpick: Prefer to keep type definitions in a type file.

widget/settings/pages/config/bar/index.ts
```diff
@@ -261,6 +261,32 @@ export const BarSettings = (): Scrollable<Gtk.Widget, Gtk.Widget> => {
...
                 }),
+                Option({
+                    opt: options.bar.workspaces.showApplicationIcons,
+                    title: 'Map Workspaces to Application Icons',
+                    subtitle: 'Use hyprctl clients class/title to match against',
+                    type: 'boolean',
+                }),
+                Option({
+                    opt: options.bar.workspaces.applicationIconOncePerWorkspace,
+                    title: 'Show the app icon only once per workspace',
```
Jas-SinghFSU, line 272:
```suggestion
                    title: 'Hide Duplicate Workspace Icons',
```

options.ts
```diff
@@ -868,6 +868,39 @@ const options = mkOptions(OPTIONS, {
...
+                obsidian: '󰠮',
+                password$: '',
+                qBittorrent$: '',
+                rofi: '',
+                slack: '',
+                spotube: '󰓇',
+                steam: '',
+                telegram: '',
+                vlc: '󰕼',
+            }),
```
Jas-SinghFSU, line 901: Prefer to keep a set of defaults in code somewhere. As this list grows and as we add more defaults, this will get unwieldy and annoying for the user to add to and modify.

We should move this to a constants file - maybe `/lib/constants/workspaces.ts` and then bring them into our workspace code. We can make it so user defined ones are concatenated to the beginning of our hard-coded defaults object so they take precedence.

This way the applicationIconMap can be empty by default.

widget/settings/pages/config/bar/index.ts
```diff
@@ -261,6 +261,32 @@ export const BarSettings = (): Scrollable<Gtk.Widget, Gtk.Widget> => {
                     title: 'Map Workspaces to Icons',
                     type: 'boolean',
                 }),
+                Option({
+                    opt: options.bar.workspaces.showApplicationIcons,
+                    title: 'Map Workspaces to Application Icons',
+                    subtitle: 'Use hyprctl clients class/title to match against',
```
Jas-SinghFSU, line 267:
```suggestion
                    subtitle: "Requires 'Map Workspace to Icons' to be enabled",
```

widget/settings/pages/config/bar/index.ts
```diff
@@ -261,6 +261,32 @@ export const BarSettings = (): Scrollable<Gtk.Widget, Gtk.Widget> => {
...
+                    type: 'boolean',
+                }),
+                Option({
+                    opt: options.bar.workspaces.applicationIconMap,
+                    title: 'App Icon Mappings',
+                    type: 'object',
+                }),
+                Option({
+                    opt: options.bar.workspaces.applicationIconFallback,
+                    title: 'Fallback App Icon',
```
Jas-SinghFSU, line 282:
```suggestion
                    title: 'Fallback App Icon',
                    subtitle: 'Fallback icon to display if no specific icon is defined for the application',
```

widget/settings/pages/config/bar/index.ts
```diff
@@ -261,6 +261,32 @@ export const BarSettings = (): Scrollable<Gtk.Widget, Gtk.Widget> => {
...
+                    type: 'boolean',
+                }),
+                Option({
+                    opt: options.bar.workspaces.applicationIconOncePerWorkspace,
+                    title: 'Show the app icon only once per workspace',
+                    type: 'boolean',
+                }),
+                Option({
+                    opt: options.bar.workspaces.applicationIconMap,
+                    title: 'App Icon Mappings',
```
Jas-SinghFSU, line 277:
```suggestion
                    title: 'App Icon Mappings',
                    subtitle: "Use the class/title output of 'hyprctl clients' to match against",
```

Jas-SinghFSU (changes requested):
Love it! It's working perfectly. There are some nitpicks I have that are minor and I would also like to move the defaults to be defined in-code; reasoning is discussed in my prior comments.

And I think there are some linter errors.

Other than that it's awesome! Thank you for the contribution, greatly appreciated :heart:.

Jas-SinghFSU:
And if we can change the defaults to:
```
Fallback Icon: 󰣆
Empty Workspace Icon: 
```

Jas-SinghFSU:
I plan to deprecate `modules/bar/workspaces/variants/default.ts` soon but we also need to pass in the missing parameters in the `renderLabel` method there.

modules/bar/workspaces/utils.ts
albohlabs: I have extracted the typings into `types/workspace.d.ts`. Please have a look at the naming if it fits your codebae.

options.ts
albohlabs: Done

Jas-SinghFSU:
> I plan to deprecate `modules/bar/workspaces/variants/default.ts` soon but we also need to pass in the missing parameters in the `renderLabel` method there.

This is the last thing we need to handle and we're good! Thank you for resolving the rest.

albohlabs:
I've added the changes to the `modules/bar/workspaces/variants/default.ts` file but didn't test it. 

I think we are done here. Thanks for guiding me. And thanks for this great repo. I ditched my own ags config in favour of it. So keep up the great work. :beers:

Jas-SinghFSU (approved):

Jas-SinghFSU:
We are good to merge! :tada: Thank you

PR #384, opened 2024-10-27
feat: raise maximum volume

rubiin:
This allows raising the speaker volume to 150% max .

This is similar to the KDE and Gnome raise maximum volume feature which is also in pavucontrol

image

<h1> screenshot </h1>

image

image


modules/menus/audio/active/SelectedPlayback.ts
```diff
@@ -35,25 +38,27 @@ const renderActivePlayback = (): Box<Child, Attribute>[] => {
...
-                            value: audio['speaker'].bind('volume'),
-                            class_name: 'menu-active-slider menu-slider playback',
-                            draw_value: false,
-                            hexpand: true,
-                            min: 0,
-                            max: 1,
-                            onChange: ({ value }) => (audio.speaker.volume = value),
-                        }),
-                    ],
+                    children: raiseMaximumVolume.bind('value').as((raiseMaxVol) => {
```
Jas-SinghFSU, line 41: Could we move this binding down to the property reliant upon it? In this case the max property, to some thing like:
```ts
                            max: raiseMaximumVolume.bind('value').as((raiseMax) => (raiseMax ? 1.5 : 1)),

```

Jas-SinghFSU: Generally I've notice stacking behavior when you put a binding inside a binding (our top level raiseMaxVol binding nesting the `audio.bind('speaker')`.

This way only the properties truly relying on the value are bounded.

Jas-SinghFSU:
This is neat! The location where the option to enable currently sits is for bar specific module. Since this is a menu item for the volume menu, I think we should put it in the `Volume Menu` section - which we'll have to create.

rubiin:
yeah. I just added this to bar as there was no menu for volume.

rubiin:
Added the menu section , also resolved the earlier comment

modules/menus/audio/active/SelectedPlayback.ts
```diff
@@ -50,7 +53,7 @@ const renderActivePlayback = (): Box<Child, Attribute>[] => {
                             draw_value: false,
                             hexpand: true,
                             min: 0,
-                            max: 1,
+                            max: raiseMaximumVolume.bind('value').as((raiseMax) => (raiseMax ? 1.5 : 1)),
                             onChange: ({ value }) => (audio.speaker.volume = value),
```
Jas-SinghFSU, line 57:
```suggestion
                            max: 1,
                            onChange: ({ value }) => (audio.speaker.volume = value),
                            setup: (self) => {
                                self.hook(raiseMaximumVolume, () => {
                                    self.max = raiseMaximumVolume.value ? 1.5 : 1;
                                });
                            },
```
For some reason the bar doesn't load on startup if we set the max value dynamically. So we have to do this approach where we set a static value and then update it in a setup hook.

PR #385, opened 2024-10-27
feat: add more lang icons

rubiin:


Jas-SinghFSU:
I don't think this code is used anymore. Maybe I should remove it.

rubiin:
@Jas-SinghFSU sure thing. Also we need to centralize icons and not use them as strings rather use them after importing them from some file . This will make future PRS more easier

PR #390, opened 2024-10-28
doc: Update README.md

rubiin:
gvfs is actually needed to cache media covers

context https://github.com/Aylur/astal/issues/52#event-14858423728

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ swww
 pacman:
 
 ```bash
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Should we add this to the 'optional' dependency section with a comment on what it's for?

rubiin:
If you are using media player, you will run into this issue. However , it can be added to main deps as its like a meg in size and  has no deps of its own

PR #413, opened 2024-10-31
feat: launcher icon auto detect from os name

rubiin:


lib/variables.ts
```diff
@@ -13,3 +13,68 @@ export const distro = {
     id: GLib.get_os_info('ID'),
     logo: GLib.get_os_info('LOGO'),
 };
+
+export const distroIcon = ((): string => {
```
Jas-SinghFSU, line 17: I like this but let's make it a 2d array as that's easier to manage and extend. Ex:
```ts
['arch', ''],
['nixos', '']

Comment (Jas-SinghFSU) [reply lib/variables.ts:17]: Then we can just do something like
```ts
const distroIcon = distroIconMap.find(distroItem => distroItem[0] === distro.id);

return distroIcon || '';

Comment (rubiin) [reply lib/variables.ts:17]: sure thing.

**Comment by Jas-SinghFSU** [comment]
Looks good but we can organize this better. Let's move the `distroIcons` array to `lib/constants/distro.ts` (new file). Then we can move the `getDistroIcon()` function into the bottom of Utils.

**Comment by rubiin** [comment]
I also updated some code to use the helper `capitalizeFirstLetter` as we were doing it manually in different places

File: lib/utils.ts [hunk 2]
```diff
@@ -11,6 +11,8 @@ import { NotificationArgs } from 'types/utils/notify';
 import { SubstituteKeys } from './types/utils';
 import { Window } from 'types/@girs/gtk-3.0/gtk-3.0.cjs';
 import { namedColors } from './constants/colors';
+import { distroIcons } from './constants/dissto';
```
Jas-SinghFSU, line 14:
```suggestion
import { distroIcons } from './constants/distro';
```

PR #426, opened 2024-11-02
feat: osd transitions

rubiin:


Jas-SinghFSU:
I'm not seeing the transitions take effect for some reason :thinking:

rubiin:
> I'm not seeing the transitions take effect for some reason 🤔

Yeah. Just checked even before the PR , if i change the transition to slide_up, I couldnt see any change. Was there an issue?

rubiin:
@Jas-SinghFSU if you have time can you look into why animations are not working with osd. The osd has crossfade now but that doesnt seem to do anything. I even changed to slide_up to get this kind of effect but no luck. Ignore this pr for now

https://github.com/System64fumo/syshud/blob/main/preview.gif

Jas-SinghFSU:
It's because we're toggling the window on and off, which interrupts the transitions. We have to do this because otherwise, the OSD will occupy the screen space when not visible (reavealed: false) and you can't interact with the window under the invisible OSD. Just some funk we have to deal with unfortunately.

rubiin:
I think I will close it then

rubiin:
@Jas-SinghFSU maybe we can pick some stuffs from here to mimic the implementation

PR #435, opened 2024-11-04
fix: display media total length on bar and add a time label

rubiin:
display total media length alongside the media position

Also a dedicated time label , ability to hide author, album details on menu and  custom text to display when nothing is playing
- Can be configured in Configuration > Media Menu

image

image


Jas-SinghFSU:
I've made some styling updates and consolidated some code for the media module. Could you please merge the branch `media_menu` into this when you get a chance? We should be good to merge afterwards.

Issue #341, opened 2024-10-20
[Feature] Send notification when battery is at a critical level


smathles:
When I get distracted on hyprland/hyprpanel, I sometimes notice "oh dang I'm on 10% battery". 

It would be nice to have a notification sent to indicate "Low power: 15% battery" and "Critical power level: 5% battery" or something along those lines like Plasma/Gnome/W*ndows does.

**Describe alternatives you've considered**
Git gud and notice when laptop is going flat lol.

anotherhadi:
There are already some good services to do that: https://github.com/electrickite/batsignal

rubiin:
I think its achievable within here without third party itself. I will try a POC

rubiin:
> There are already some good services to do that: [electrickite/batsignal](https://github.com/electrickite/batsignal)

I second this. I think its better to use this however the above mention DIY works well too

smathles:
True, there are several ways of getting this functionality that I've seen (ags services, systemd services, bash script that's always running in the background. I was mainly thinking of having this service automatically a part of the ags system running, as we already have that going and we have battery/notifications/power profiles already set up.

I don't think it would be too difficult to take the existing battery/power profiles/notifications and make a service in ags that does this automatically. Maybe batsignal etc is more minimal and more functional, but it's nice to have a bar that does "everything", instead of having to build several things yourself still (I get that's not what everyone's looking for, but imo then probably hyprpanel isn't the most minimal solution either).

rubiin:
@smathles could you check the barnch ?

smathles:
Unfortunately am visiting family, where there is no internet ;).

Remind me in a few days when I get reception again haha.

Surely slide into the discord server if you're not already?

rubiin:
No worries

smathles:
@rubiin It looks like I can't actually test this, as it's only available on your repo. Unless it's a branch on Jas's repo, I can't test it (unless I feel like being gigglish and adding a third upstream source to my single repo loll).

Jas-SinghFSU:
@smathles You can add this into your `~/.zshrc`
```ts
function gpr() {
    if [ -z "$1" ]; then
        echo "Usage: gpr <pull-request-number>"
        return 1
    fi
    git fetch origin pull/$1/head:pr-$1 && git checkout pr-$1
}
```

Then to check out a pull request you just do `gpr pullRequest#`... So:
```sh
gpr 441
```

smathles:
[...]

PR #441, opened 2024-11-05
fix: add battery notif

rubiin:
closes https://github.com/Jas-SinghFSU/HyprPanel/issues/341


lib/utils.ts
```diff
@@ -193,3 +195,23 @@ export const isValidGjsColor = (color: string): boolean => {
...
+                    `${battery.percent}% Battery Percentage`,
+                    '-i',
+                    icons.battery.warning,
+                    '-u',
+                    'critical',
+                ]);
+            }
+        });
+    });
+}
```
Jas-SinghFSU, line 217: You shouldn't need the top level bindings here since the values being checked will be at the moment that the event triggers. So if it triggers at 20%, then at that moment it will check the values as they are, so no need to bind them.

```suggestion
export function warnOnLowBattery(): void {
    battery.connect('notify::percent', () => {
        const { lowBatteryThreshold, lowBatteryNotification } = options.menus.power;

        if (!lowBatteryNotification.value || battery.charging) return;

        const lowThreshold = lowBatteryThreshold.value;

        if (battery.percent === lowThreshold || battery.percent === lowThreshold / 2) {
            Utils.execAsync([
                'notify-send',
                `${battery.percent}% Battery Percentage`,
                '-i',
                icons.battery.warning,
                '-u',
                'critical',
            ]);
        }
    });
}

```

Jas-SinghFSU, line 217: You shouldn't need the top level bindings here since the values being checked will be at the moment that the event triggers. So if it triggers at 20%, then at that moment it will check the values as they are, so no need to bind them.

```suggestion
export function warnOnLowBattery(): void {
    battery.connect('notify::percent', () => {
        const { lowBatteryThreshold, lowBatteryNotification } = options.menus.power;

        if (!lowBatteryNotification.value || battery.charging) return;

        const lowThreshold = lowBatteryThreshold.value;

        if (battery.percent === lowThreshold || battery.percent === lowThreshold / 2) {
            Utils.execAsync([
                'notify-send',
                `${battery.percent}% Battery Percentage`,
                '-i',
                icons.battery.warning,
                '-u',
                'critical',
            ]);
        }
    });
}
```

rubiin: will this be reactive. When I update the from the settings will it be reflected back on the current ags session or have to restart

Jas-SinghFSU: It will be because every time the battery.connect event triggers, it'll fetch the value at that point on time. All these values are, are just variables at the end of the day.

So if you update the value in the settings, the next time the event triggers after your settings change, it will fetch the variables value which will be the new value.

rubiin: Oh yeah since its tied to event its supposed to be reactive.

lib/utils.ts
```diff
@@ -196,6 +198,23 @@ export const capitalizeFirstLetter = (str: string): string => {
...
+        if (!lowBatteryNotification.value || battery.charging) return;
+
+        const lowThreshold = lowBatteryThreshold.value;
+
+        const notifText = lowBatteryNotificationText.value.replace(/\$POWER_LEVEL/g, `${lowThreshold}%`);
+
+        if (battery.percent === lowThreshold || battery.percent === lowThreshold / 2) {
+            Utils.execAsync(['bash', '-c', `notify-send 'Warning: Low battery ' '${notifText}' -u critical`]).catch(
+                print,
+            );
```
Jas-SinghFSU, line 214:
```suggestion
            Notify({
                summary: 'Warning: Low battery',
                body: notifText,
                iconName: icons.ui.warning,
                urgency: 'critical',
                timeout: 7000,
        });

```

lib/utils.ts
```diff
@@ -11,6 +11,8 @@ import { NotificationArgs } from 'types/utils/notify';
 import { SubstituteKeys } from './types/utils';
 import { Window } from 'types/@girs/gtk-3.0/gtk-3.0.cjs';
 import { namedColors } from './constants/colors';
+const battery = await Service.import('battery');
+import options from 'options';
```
Jas-SinghFSU, line 15:
```suggestion
import options from 'options';
import { Notify } from 'lib/utils';
```

widget/settings/pages/config/menus/power.ts
smathles, line 1: remove double space between "Low" and "Battery" in line 29

options.ts
smathles, line 1: No space between ($POWER_LEVEL) and next sentence. I would recommend adding some \n's, so it looks better.

image


`Your battery is running low ($POWER_LEVEL).\n\nPlease plug in your charger.`

Could also get rid of the whole "Your battery is running low" sentence and just add the percentage charge to the title. Purely a stylistic preference, you do you :)

widget/settings/pages/config/menus/power.ts
rubiin: double space is due to icon so I think it can't be removed

smathles: Which icon? I think it's just text, unless I'm confusing something.

rubiin: the warning icon

rubiin: image

options.ts
rubiin: might allow a custom text for header too.

widget/settings/pages/config/menus/power.ts
smathles: Ah, we're looking at different things.

I mean this line here.

image

smathles: image

I think?

rubiin: it was moved I guess

options.ts
```diff
@@ -1120,8 +1120,9 @@ const options = mkOptions(OPTIONS, {
             showLabel: opt(true),
             lowBatteryThreshold: opt(20),
             lowBatteryNotification: opt(false),
-            lowBatteryNotificationText: opt(
-                'Your battery is running low ($POWER_LEVEL).\n\nPlease plug in your charger.',
+            lowBatteryNotificationText: opt('Warning: Low battery'),
+            lowBatteryNotificationTitle: opt(
+                'Your battery is running low ($POWER_LEVEL %).\n\nPlease plug in your charger.',
```
smathles, line 1125: I think the default title should be something like `Warning: Low Battery ($POWER_LEVEL%)` and the default text something like `Please plug in your PC`. Both of them being the same message looks a bit strange atm.

rubiin: takeing reference from some solutions like this

smathles: looks good

DivineMK:
The $POWER_LEVEL is not replaced for me

PR #453, opened 2024-11-08
fix: wifi menu scrollable

rubiin:
before the pr, the wifi menu would stretch which would could cause the menu to fill the entire vertical width if there were 15-20 + wifis
image

The menu is made a scrollable so it has a constant height irrespective of how many networks are available
[Kooha-2024-11-08-09-07-42.webm](https://github.com/user-attachments/assets/648fff46-606e-47a0-a0a6-85c18e0ae43d)

### DIFF
--- a/modules/menus/network/wifi/index.ts
+++ b/modules/menus/network/wifi/index.ts
@@ -56,7 +56,7 @@ const Wifi = (): BoxWidget => {
                 ],
             }),
             Widget.Scrollable({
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Can we add scrollbar styling similar to how it is in the settings dialog (maybe thinner)? That way users know when hovering over the menu that there are more elements to scroll. We may need to do the same for bluetooth.

PR #480, opened 2024-11-11
fix: line appearing on notification panel bottom

rubiin:
image

Removes this black line from appearing


Jas-SinghFSU:
If we remove this, we must also remove the setting in options to configure the background color as it's no longer used. And the settings menu entry in widget/settings/pages/theme/menus/notifications.ts.

Jas-SinghFSU:
I also think there is a better way to handle it than to rip it out.

rubiin:
gonna close this

PR #502, opened 2024-11-14
fix: idle inhibitor

rubiin:
closes https://github.com/Jas-SinghFSU/HyprPanel/discussions/496


dotaxis:
I've tested this and it works great!

For some reason the styling options seem to be missing on my end, so I can't change the colour of the module.

I also think the default icons should be swapped (so the open eye means idling is inhibited), but they're configurable so that's maybe just me being pedantic.

Thank you very much for this!

sitolam:
I can't find this feature am I missing something?

PR #505, opened 2024-11-15
feat(media): truncate media label with consideration of full-width characters

sakarie9:
Previously, all characters were treated as 1 unit wide, which caused incorrect truncation for certain characters, such as CJK characters. This resulted in media labels being longer than expected when compared to ASCII characters.

For instance, when the "Media - Truncation Size" is set to 30, the truncation works correctly for alphabetic characters, displaying up to 30 characters. However, for CJK characters, each character was counted as 1 unit, causing the truncation to display approximately twice the width compared to 30 alphabetic characters.

This commit introduces a Regex solution to count full-width characters as 2 units in length. While this approach may not cover every possible full-width character, it should be sufficient for most common scenarios.

### DIFF
--- a/modules/bar/media/helpers.ts
+++ b/modules/bar/media/helpers.ts
@@ -39,6 +39,30 @@ const isValidMediaTag = (tag: unknown): tag is keyof MediaTags => {
     return (mediaTagKeys as readonly string[]).includes(tag);
 };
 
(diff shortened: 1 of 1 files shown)


rubiin (approved):

Jas-SinghFSU:
Hey @sakarie9, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #576, opened 2024-12-21
Fix weather disable not working

### LINKED ISSUE #578: Clock module (Time & Weather) doesn't respond to settings changes
**Describe the bug**
The clock module settings isn't responding by any change in the settings.

**To Reproduce**
I wanted to hide the seconds and use the 24Hr time mode, but whatever I change in the hyprpanel settings in the clock section it doesn't affect at all, even the weather settings 

image
 - Distribution: Arch Linux(EOS)

**Additional context**
note that this bug was even before migrating to astal/ags2, I didn't report waited to the migration release.

davfsa:
During AGS v2 migration, this seems to have been missed, so the weather widget would always display, no matter the value of the setting

Closes #578


Jas-SinghFSU:
Thanks! :heart:

PR #577, opened 2024-12-21
Properly implement low battery notification service

### LINKED ISSUE #575: Low battery notification is not sent
**Describe the bug**
The low battery notification is not sent when battery drops bellow the limit

**To Reproduce**
1. Enable low battery notification
image
2. Wait for the battery to drop under that level
3. Notice no notification is sent

There are quite bad reproduction steps, but if needed I can gather and post more information :)

**Expected behavior**
Notification is sent when battery drops under threshold.

**Screenshots**
N/A

**Desktop:**
 - Distribution: Arch Linux

**Additional context**
This happened before ags v2 migration, but I waited to report it until the migration was complete. It unfortunately still happens

davfsa:
The original implementation had multiple issues:
1. The service itself would never trigger as it listened to be notified of the wrong variable. Should have been `percentage` not `percent`
2. `.percentage` is reported as a double between 0 and 1, so setting the threshold to a percentage representation of the threshold (20 for example) would never trigger it
3. Strict comparison of doubles would never work as the battery percentage will always be way more precise than 2 decimal points
4. The replacement of the magic value `$POWER_LEVEL` did not work at all

Instead I opted for an implementation that checks for the values bellow the threshold and send the notification if bellow that value, resetting this when the battery is reported as charging

Closes #575

### DIFF
--- a/src/lib/behaviors/batteryWarning.ts
+++ b/src/lib/behaviors/batteryWarning.ts
@@ -3,7 +3,18 @@ import icons from '../icons/icons';
 import { Notify } from '../utils';
 
 export function warnOnLowBattery(): void {
-    batteryService.connect('notify::percent', () => {
+    var sentLowNotification = false;
+    var sentHalfLowNotification = false;
+
+    batteryService.connect('notify::charging', () => {
+        // Reset it when the battery is put to charge
+        if (batteryService.charging) {
+            sentLowNotification = false;
+            sentHalfLowNotification = false;
+        }
+    });
+
+    batteryService.connect('notify::percentage', () => {
         const { lowBatteryThreshold, lowBatteryNotification, lowBatteryNotificationText, lowBatteryNotificationTitle } =
             options.menus.power;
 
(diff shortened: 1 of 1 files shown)


src/lib/behaviors/batteryWarning.ts
```diff
@@ -3,22 +3,45 @@ import icons from '../icons/icons';
 import { Notify } from '../utils';
 
 export function warnOnLowBattery(): void {
-    batteryService.connect('notify::percent', () => {
+    var sentLowNotification = false;
+    var sentHalfLowNotification = false;
```
Jas-SinghFSU, line 7:
```suggestion
    let sentHalfLowNotification = false;
```

src/lib/behaviors/batteryWarning.ts
```diff
@@ -3,22 +3,45 @@ import icons from '../icons/icons';
...
         }
 
+        // batteryService.percentage will be a double in between 0 and 1, so we multiply it by 100 to convert it to the percentage
+        const batteryPercentage = Math.floor(batteryService.percentage * 100);
         const lowThreshold = lowBatteryThreshold.get();
 
-        if (batteryService.percentage === lowThreshold || batteryService.percentage === lowThreshold / 2) {
+        // To avoid double notifications, we check each of the thresholds and set the correct `sentNotification`, but then
+        // combine them into one notification only
+        var sendNotification = false;
```
Jas-SinghFSU, line 31:
```suggestion
        let sendNotification = false;
```

Jas-SinghFSU:
Ty :heart:

PR #579, opened 2024-12-21
Fix: option mapping for bluetooth bar module label

okt:
`bar.bluetooth.label` was previously mapped to `bar.volume.label` incorrectly in the bar's bluetooth module. The options UI was not allowing the toggle to work as intended.


Jas-SinghFSU:
Ty :heart:

PR #587, opened 2024-12-22
Upgraded the Nix install to work with Astal

ThomasCarey4:
You forgot to change the nix installation for Astal lol, was completely broken.

I cleaned up the flake.nix to use the inbuilt ags bundler (instead of bun) and had to modify a few ts imports as the bundler gets confused by the same file being imported by different paths.


ThomasCarey4:
Whoops, must of missed them thanks!

Jas-SinghFSU:
The file path issue seems to be nix specific.

Jas-SinghFSU:
We should figure out if there is a way to handle that properly first.

Jas-SinghFSU:
Actually I think the import updates are fine but the ts config path resolution seems redundant. That shouldn't be needed?

ThomasCarey4:
> Actually I think the import updates are fine but the ts config path resolution seems redundant. That shouldn't be needed?

yeah ur right the ts config change does nothing lol. Never used typescript so I was just asking chatgpt for it all :)

tsconfig.json
```diff
@@ -21,6 +21,7 @@
         "jsx": "react-jsx",
         "jsxImportSource": "/usr/share/astal/gjs/gtk3",
         "paths": {
+            "src/*": ["src/*"],
```
Jas-SinghFSU, line 24:
```suggestion
```

PR #592, opened 2024-12-22
Fix: Matugen Global themes

paridhi089:
### Description
with the addition of ` --dry ` flag to the matugen command in the past update, Matugen only generates themes for Hyprpanel. 

If the user has other configs for other applications, the necessary changes to the colour scheme wont be made for them. The user can run the matugen command again to rectify this. But this is not ideal. 

With the changes i have made, if the user run 

```
hyprpanel setWallpaper <path>

``` 

it will change the theme for hyprpanel as well as any other applications specified in matugen config file.  Additionally changing the wallpaper from the setting panel will now also update the colorscheme on not just hyprpanel but any other apps specified in the matugen config file.

### Documentation update

In the matugen section of the documentation, users must me informed to use the below command or the settings panel to change their background instead of using ` matugen image ` .

```
hyprpanel setWallpaper <path>

``` 

Additionally it is advised that the user use matugen to only generate colourscheme and not set wallpaper.


src/services/matugen/index.ts
```diff
@@ -33,6 +33,13 @@ export async function generateMatugenColors(): Promise<MatugenColors | undefined
         const contents = await bash(
             `matugen image --dry-run -q ${wallpaperPath} -t scheme-${scheme_type.get()} --contrast ${normalizedContrast} --json hex`,
         );
+        const themeStatus= await bash(
+            `matugen image -q ${wallpaperPath} -t scheme-${scheme_type.get()} --contrast ${normalizedContrast}`,
+        );
+
+        
+        
+
```
Jas-SinghFSU, line 42:
```suggestion
```
Extra empty spaces not needed

src/services/matugen/index.ts
```diff
@@ -33,6 +33,13 @@ export async function generateMatugenColors(): Promise<MatugenColors | undefined
         const contents = await bash(
             `matugen image --dry-run -q ${wallpaperPath} -t scheme-${scheme_type.get()} --contrast ${normalizedContrast} --json hex`,
         );
+        const themeStatus= await bash(
```
Jas-SinghFSU, line 36:
```suggestion
        await bash(
```
We can remove the constant since it's not used anywhere.

src/services/matugen/index.ts
```diff
@@ -33,6 +33,13 @@ export async function generateMatugenColors(): Promise<MatugenColors | undefined
...
+            `matugen image -q ${wallpaperPath} -t scheme-${scheme_type.get()} --contrast ${normalizedContrast}`,
+        );
```
Jas-SinghFSU, line 38:
```suggestion
        await bash(`matugen image -q ${wallpaperPath} -t scheme-${scheme_type.get()} --contrast ${normalizedContrast}`);
```

PR #601, opened 2024-12-23
support paru in `checkUpdates` & clarify polling interval is in ms

amaanq:
Hey :wave: just wanted to add a few changes I thought might be nice to have.

1. The first commit changes the `checkUpdates` script to support using [`paru`](https://github.com/Morganamilo/paru) as the aur helper when fetching the # of aur updates. If the command paru exists, then that's used as the aur helper, else by default yay is used.
2. When looking through the config options, namely for updates, I noticed the option for the polling interval didn't clarify that the value is in `ms` -  though it might seem "obvious" that it is, I think denoting this would be helpful nonetheless.

Thanks! I'm really enjoying this bar :slightly_smiling_face:


Jas-SinghFSU:
Ty :smile:

PR #497, opened 2024-11-13
feat: allow reading of JSON objects in the configuration

dsluijk:
This PR allows HyprPanel to read conventional JSON objects from the configuration file, while keeping the dot-notation as used currently. This means it should not break existing workflows. The behaviour with writing to the configuration file is not changed.

Should also fix #433, as I was running into the same issue.

PR #605, opened 2024-12-23
Add Home Manager module

benvonh:
Added `module.nix` to the `nix` folder.

README is updated to show how to use it.

A few design problems are:
1. There is no clean way to reload the config. You have to stop HyprPanel and restart it.
2. Waiting for https://github.com/Jas-SinghFSU/HyprPanel/pull/497 to be merged so that the config generation logic can be drastically simplified.
3. Overwriting `config.json` through the GUI is a problem due to how Nix works. The best workaround is to generate a file highlighting the differences instead of overwriting it so that users can easily update their config in the Nix module. This has to be done manually by nature but would also be very nice to have (low priority).


benvonh:
@orangci I replaced it with a link to the installation page.

Also, the module was made to handle without that PR but I'll set this as a draft anyway.

If anyone wants to try using it now, just change the flake input URL to my fork.

benvonh:
Yes, it's ready - I wouldn't submit a PR if it wasn't xd

It's just that those design problems make the inside logic messy. But functionality-wise, it's all working. I'll slowly fix them at a later date.

PR #606, opened 2024-12-23
Fix resolving paths for .face.icon in dashboard and add home ('~') path support

davfsa:
I noticed that my `.face.icon` was not being loaded and, after some investigation, I uncovered that `GdkPixbuf.Pixbuf` doesn't expand paths, let alone environment variables (which the default config was attempting to use through `$HOME`).

This fixes it by adding support for `~` to `isAnImage`, making use of it in avatar resolution and resolving the path to load the avatar.

---

I have also simplified `isAnImage` by removing the double check for file existence, as `GdkPixbuf.Pixbuf.new_from_file` does it itself


src/lib/utils.ts
```diff
@@ -228,25 +227,39 @@ export function launchApp(app: AstalApps.Application): void {
...
+ * Note: This will only expand '~' if present. Path traversal is not supported.
+ *
+ * @param path The path to resolve.
+ *
+ * @returns The absolute representation of the resolved path.
+ */
+export function resolvePath(path: string): string {
+    if (path.charAt(0) == '~') {
+        path = GLib.get_home_dir() + '/' + path.substring(1, path.length);
+    }
```
davfsa, line 258:
```suggestion
    return path.replace('~', GLib.get_home_dir())
```

Might be a bit better

src/lib/utils.ts
```diff
@@ -228,25 +227,37 @@ export function launchApp(app: AstalApps.Application): void {
...
+ * Note: This will only expand '~' if present. Path traversal is not supported.
+ *
+ * @param path The path to resolve.
+ *
+ * @returns The absolute representation of the resolved path.
+ */
+export function resolvePath(path: string): string {
+    return path.replace('~', GLib.get_home_dir())
+
+    return path;
```
Jas-SinghFSU, line 258:
```suggestion
```

src/lib/utils.ts
```diff
@@ -228,25 +227,35 @@ export function launchApp(app: AstalApps.Application): void {
...
+/**
+ * Resolve a path to the absolute representation of the path.
+ *
+ * Note: This will only expand '~' if present. Path traversal is not supported.
+ *
+ * @param path The path to resolve.
+ *
+ * @returns The absolute representation of the resolved path.
+ */
+export function resolvePath(path: string): string {
```
Jas-SinghFSU, line 255: We should probably name this `normalizePath` as it simply makes it absolute and doesn't quite resolve a path.

davfsa: Sounds good!

Jas-SinghFSU:
Thanks! :smile:

PR #613, opened 2024-12-24
fix: missing class for bluetooth discover button

Diaoul:
The discover button is not styled correctly, I'm guessing this is because the bluetooth class is missing, I did not test though.

image

Based off [this styling](https://github.com/Jas-SinghFSU/HyprPanel/blob/master/src/components/menus/bluetooth/header/Controls/ToggleSwitch.tsx#L13) that is working and [this one](https://github.com/Jas-SinghFSU/HyprPanel/blob/master/src/components/menus/network/wifi/Controls/RefreshButton.tsx#L10) from the network menu.

```html
className="menu-icon-button search network"
```


Diaoul:
Confirmed it's fixed
image

PR #630, opened 2024-12-26
feat: add flatpak update count

ilikestreet:
Really simple this PR,

- Get flatpak update count in update script

### DIFF
--- a/scripts/checkUpdates.sh
+++ b/scripts/checkUpdates.sh
@@ -1,26 +1,36 @@
 #!/bin/bash
 
+check_flatpak_updates() {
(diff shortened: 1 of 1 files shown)


ilikestreet:
For ppl dont have flatpak, It wont affect there update count. Otherwise it shows up in the total count.

Jas-SinghFSU:
Flatpak is really terrible at being consistent in it's results. I think this was a valiant effort but even in my usage I've noticed that the regex isn't robust enough to always capture the update count properly since the results can come back differently in many situations. Closing for that reason until Flatpak provides a better update count :(

PR #646, opened 2024-12-28
nix: Add overlay to HM module + useful script

benvonh:
The "useful script" just does a `colordiff` on `config.json` with a *swap* version.

Play around from within the GUI and do `hyprpanel-diff` to see what changes you made from the GUI.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -7,7 +7,7 @@ let
   jsonFormat = pkgs.formats.json { };
   
   # No package option
-  package = self.packages.${pkgs.system}.default;
+  package = pkgs.hyprpanel;
 
   # Shorthand lambda for self-documenting options under settings
   mkStrOption = default: mkOption { type = types.str; default = default; };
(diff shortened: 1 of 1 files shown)


zenitraM:
This makes the overlay not compatible with this homeManager option fwiw:

```
Failed assertions:
       - zenitram profile: `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
```
see https://discourse.nixos.org/t/home-manager-useuserpackages-useglobalpkgs-settings/34506.

Maybe a separate overlay to be added to nixpkgs could be provided for those cases?

benvonh:
@zenitraM not sure about separate overlay but I assume you meant a NixOS module. I don't think it's worth adding a whole module just to overlay a package so I left it as an option instead. `overlay.enable = true` to have the HM module overlay for you; or simply add the overlay yourself for people using global packages. (im using it now too lol thanks) https://github.com/Jas-SinghFSU/HyprPanel/pull/654

Issue #381, opened 2024-10-27
Finding and editing te right json files


ShinobiX9X:
In the Configuration tool, some settings allow adding json style text in a small box.
![image](https://github.com/user-attachments/assets/83e48a11-5d47-4503-9d22-49e3ac8e381c)

It would be nice to have the pwd of that particular json file showed to know where to edit.
Or clicking a button, having the file opened in the default editor to change the file and save.

I just installed hyprpanel yesterday and love it. In seconds, I have the look I tried for long with waybar.
In some instances, especially when you want to add more workspace icons/names also in window titles, it seems
impossible to add the examples shown in the hyprpanel in the small boxes in the configuration tool.

Like this, how do you edit this in the config tool?
![image](https://github.com/user-attachments/assets/b343e0ea-e691-45e8-a3cf-cf8be6138642)

I followed installation and configuration guide, but did not read where I could do this.

Jas-SinghFSU:
The best way right now is it copy it into a real editor and paste it back in the text box when you're done; until I find a better solution. Editing the config in the text boxes is nearly impossible.

smathles:
Would love to see this implemented. Not a massively vital task maybe, but definitely improved QOL.

PR #684, opened 2024-12-31
Add live reloading of configuration file

### LINKED ISSUE #787: Hyprpanel freezing for a long time after changing themes from CLI
**Describe the bug**
Whenever there is a notification and  I try to change theme of panel using `  hyprpanel useTheme *` the panel freezes for 10-15 seconds before actually changing themes

**To Reproduce**
Send a notification and change theme when notification is showing ( it's random tho )

**Expected behavior**
HyprPanel doesn't freeze when changing themes

**Screenshots**
None

 - Distribution: EndeavourOS
- Window Manager : Hyprland

**Additional context**
None

davfsa:
This also removes the need for a file with all the available configuration and a shadow configuration file.

Additionally, added several improvements:
1. Reduce I/O on initial configuration loading by only reading file once, instead of once per option (scary! Hyprpanel has a lot of options)
2. Remove unnecesary back and forth events when editing configuration

---

There were also old references to a "cache file", which is no longer a thing since the Astal migration, so renamed those too

---

Closes #787 - The slowdown is caused by the massive I/O, which this pr fixed. There will still be a delay because they are not updated in bulk, but will be wayyyyy less


davfsa:
Kind of closes https://github.com/Jas-SinghFSU/HyprPanel/issues/381, as it is now way nicer to edit these types of configuration, but im not sure :)

Jas-SinghFSU:
Nice! I'll review this over the weekend when I get more time.

src/lib/option.ts
```diff
@@ -203,66 +177,71 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
+    const rawConfig = readFile(CONFIG);
 
-    const values: Record<string, unknown> = {};
-    const defaultValues: Record<string, unknown> = {};
+    let config: Record<string, unknown> = {};
+    if (rawConfig && rawConfig.trim() !== '') {
+        try {
+            config = JSON.parse(rawConfig) as Record<string, unknown>;
+        } catch (error) {
+            errorHandler(error);
```
davfsa, line 195: Should also send a notification if the initial load failed

src/lib/option.ts
```diff
@@ -203,66 +177,71 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
         }
 
         for (let i = 0; i < allOptions.length; i++) {
             const opt = allOptions[i];
             const newVal = cache[opt.id];
-            const oldVal = opt.get();
 
+            if (newVal === undefined) {
+                continue;
+            }
```
davfsa, line 238: In this case, we would want to reset the value to the default

src/lib/option.ts
```diff
@@ -203,66 +177,71 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
+            // 200 milliseconds event debounce
+            return;
+        }
+        lastEventTime = Date.now();
 
-    monitorFile(configFile, () => {
-        const raw = readFile(configFile);
+        const raw = readFile(CONFIG);
 
         if (!raw || raw.trim() === '') return;
```
davfsa, line 217: Here we would just want to treat it as an empty config and reload it as such

src/lib/option.ts
```diff
@@ -76,118 +73,106 @@ export class Opt<T = unknown> extends Variable<T> {
...
-
-        let defaultData: Record<string, unknown> = {};
-
-        if (rawData && rawData.trim() !== '') {
-            try {
-                defaultData = JSON.parse(rawData) as Record<string, unknown>;
-            } catch {
-                // do nuffin
-            }
+    public set(value: T, { writeDisk = true }: WriteDiskProps = {}): void {
```
Jas-SinghFSU, line 94:
```suggestion
    public set = (value: T, { writeDisk = true }: WriteDiskProps = {}): void => {
```
Because Variable<T> is wrapped in a Proxy, its methods are treated as properties holding functions. Let's get typescript to stop whining by redefining `set` as an arrow function so it matches the base type.

src/lib/option.ts
```diff
@@ -76,118 +73,106 @@ export class Opt<T = unknown> extends Variable<T> {
...
+        if (value == this.get()) {
```
Jas-SinghFSU, line 95:
```suggestion
        if (value === this.get()) {
```

src/lib/option.ts
```diff
@@ -76,118 +73,106 @@ export class Opt<T = unknown> extends Variable<T> {
...
+    public reset(writeDiskProps: WriteDiskProps = {}): string | undefined {
         if (this.persistent) {
             return undefined;
         }
 
         const current = this.get();
 
-        if (JSON.stringify(current) !== JSON.stringify(this.initial)) {
-            this.set(this.initial);
+        if (current !== this.initial) {
```
Jas-SinghFSU, line 139:
```suggestion
        let currentValue: string | T = this.get();
        currentValue = typeof this.get() === 'object' ? JSON.stringify(currentValue) : currentValue;

        let initialValue: string | T = this.get();
        initialValue = typeof this.get() === 'object' ? JSON.stringify(initialValue) : initialValue;

        if (currentValue !== initialValue) {
```
This condition will always evaluate to false for objects. Let's stringify if the value is an object and then compare. Otherwise just compare the value as is if not object.

src/lib/option.ts
```diff
@@ -76,118 +73,106 @@ export class Opt<T = unknown> extends Variable<T> {
...
+     * @param writeDisk - Whether to write the changes to disk. Defaults to true.
      * @returns Returns the option's ID if reset occurred, otherwise undefined.
      */
-    public reset(): string | undefined {
+    public reset(writeDiskProps: WriteDiskProps = {}): string | undefined {
         if (this.persistent) {
             return undefined;
         }
 
         const current = this.get();
```
Jas-SinghFSU, line 137:
```suggestion
```
Remove in favor of suggestion below.

src/lib/option.ts
```diff
@@ -233,66 +218,80 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
-
-    for (let i = 0; i < allOptions.length; i++) {
-        const option = allOptions[i];
-        const val = option.value;
+export function mkOptions<T extends object>(optionsObj: T): T & MkOptionsResult {
+    // Ensure the directory exists
+    ensureDirectory(CONFIG.split('/').slice(0, -1).join('/'));
 
-        values[option.id] = val;
+    // Read the configuration file
```
Jas-SinghFSU, line 228:
```suggestion
```
Noisy comment

src/lib/option.ts
```diff
@@ -233,66 +218,80 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
-    const configFile = `${TMP}/${confFile}`;
-
-    const values: Record<string, unknown> = {};
-    const defaultValues: Record<string, unknown> = {};
-
-    for (let i = 0; i < allOptions.length; i++) {
-        const option = allOptions[i];
-        const val = option.value;
+export function mkOptions<T extends object>(optionsObj: T): T & MkOptionsResult {
+    // Ensure the directory exists
```
Jas-SinghFSU, line 225:
```suggestion
```
Noisy comment

src/lib/option.ts
```diff
@@ -233,66 +218,80 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
+    for (let i = 0; i < allOptions.length; i++) {
+        allOptions[i].init(config);
+    }
 
-        if (!raw || raw.trim() === '') return;
+    // Setup a file monitor to allow live config edit preview from outside
+    // the config menu
+    let lastEventTime = Date.now();
+    monitorFile(CONFIG, () => {
+        if (Date.now() - lastEventTime < 200) {
```
Jas-SinghFSU, line 256:
```suggestion
        const debounceTime = 200;
        const timeSinceLastEvent = Date.now() - lastEventTime;

        if (timeSinceLastEvent < debounceTime) {
```

src/lib/option.ts
```diff
@@ -233,66 +218,80 @@ function getOptions(object: Record<string, unknown>, path = '', arr: Opt[] = [])
...
+            // 200 milliseconds event debounce
```
Jas-SinghFSU, line 257:
```suggestion
```

Jas-SinghFSU:
Looks good to me other than the comments :+1:

Jas-SinghFSU:
> Nice! I'll review this over the weekend when I get more time.

The weekend came 2.5 months late lol

src/lib/option.ts
davfsa: This seems to make typescript complain even more https://github.com/Jas-SinghFSU/HyprPanel/actions/runs/13882049556/job/38841835707?pr=684 :thinking:

src/lib/option.ts
```diff
@@ -76,118 +73,109 @@ export class Opt<T = unknown> extends Variable<T> {
...
      * @returns Returns the option's ID if reset occurred, otherwise undefined.
      */
-    public reset(): string | undefined {
+    public reset(writeDiskProps: WriteDiskProps = {}): string | undefined {
         if (this.persistent) {
             return undefined;
         }
 
-        const current = this.get();
+        const currentValue: string | T = this.get();
```
Jas-SinghFSU, line 137:
```suggestion
        let currentValue: string | T = this.get();
```

davfsa: Whoops!

Nice catch!

PR #690, opened 2025-01-03
feat: add timezone

noahfraiture:
Hello,

I would like to make multiple timezone possibility for the clock. The code is a tentative to simply split in the first time but this doesn't work because format isn't the string itself. I don't except it to work that way but I think it can be a good start of discussion.

How would you see it ?


Jas-SinghFSU:
#795 iterates on this.

PR #702, opened 2025-01-08
feat: add padding option for inactive workspace numbers

keonly:
Hi, I added an extra option to enable setting the paddings for inactive numbered workspaces.

- Before
snapshot_2025-01-09_05-19-20
snapshot_2025-01-09_05-19-36
(Notice the inconsistency of the workspace number positions!)
- After (`theme.bar.buttons.workspaces.numbered_inactive_padding = "0.4em";`)
snapshot_2025-01-09_05-18-47
snapshot_2025-01-09_05-21-37

In both cases, I have set `theme.bar.buttons.workspaces.numbered_active_highlight_padding` to `0.4em`. Please let me know what you think. Thanks for the great project!


keonly:
I found that the positions of the workspace buttons were still inconsistent when only `bar.workspaces.spacing` was adjusted.

Here is an example of such inconsistency when `bar.workspaces.spacing` is set to `2` and `theme.bar.buttons.workspaces.numbered_active_highlight_padding` is set to `0.4em`.

snapshot_2025-01-11_16-12-28
snapshot_2025-01-11_16-12-42

keonly:
Here are better visualisations of the change made in this PR!

- Before (adjusted with spacing only)

https://github.com/user-attachments/assets/d3a2e9b2-f666-44ae-afd5-c11f224bdad8
- After

https://github.com/user-attachments/assets/f480bdd8-7810-4538-ad18-3d034bc4b138

Jas-SinghFSU:
Nice, in that case we should probably remove the spacing option since this is more robust.

Jas-SinghFSU:
Can we move the Inactive Padding setting righ below the Highlight Padding so it's logically grouped.

keonly:
Updated! Would this be good?

Jas-SinghFSU:
Yep, thanks for the contribution :heart:

PR #709, opened 2025-01-10
Add up-to-date default workspace labels

davfsa:
They now are built from `defaultWindowTitleMap`, which has also been extracted to avoid continuous instantiation, which should aid with memory usage.

---

Additionally, I have also bundled into this pull request (but as a separate commit in case a different pr is better or the change is just not wanted) a patch to use iterators instead of always copying the mappings for the icons (workspaces + windows), which should further lower memory usage and provide optimizations for large mappings


davfsa:
A followup to this pr would be to add some new workspace labels to identify some more common applications, but wanted to get the bases for this laid first :)

src/components/bar/modules/workspaces/helpers/utils.ts
```diff
@@ -39,18 +39,15 @@ const isWorkspaceActiveOnMonitor = (monitor: number, i: number): boolean => {
...
-    if (!iconEntry) {
-        return `${i}`;
-    }
-
-    const hasIcon = typeof iconEntry === 'object' && 'icon' in iconEntry && iconEntry.icon !== '';
-
-    if (typeof iconEntry === 'string' && iconEntry !== '') {
-        return iconEntry;
-    }
+    if (iconEntry) {
```
Jas-SinghFSU, line 42: What's the reason we're moving away from our [guard clause](https://deviq.com/design-patterns/guard-clause)?

davfsa: I usually highly advocate for guard clauses, but in this case I preferred nesting because it's not too much and it allows to have the default return value be common between all branches and utilize fall through for it, instead of having to duplicate it.

It's a simple default, so I can revert it if it's preferred

Jas-SinghFSU: I can revert it in the branch I'm working on and make it a bit more clear. This is mostly a nitpick and a preference thing but I tend to prefer verbosity because in the prior implementation, it's more visibly clear that we return a default if:
1. iconEntry is null
2. None of the other conditions were satisfied

Once again, this is extremely minor and a preference thing so I can revert it np.

davfsa: Updated!

Jas-SinghFSU:
Firstly, consolidating the icons as you've done is phenomenal. That is something I had wanted to do but never made it on the radar, so thank you!

Second, while I agree that iterators would prevent object copies and hence memory usage, it's important to note that these object copies are ephemeral and garbage collected. The difference in memory usage would be unnoticeable however what it does introduce is more complexity into the code-base - having to manage  the iterators yourself and conditional looping exits. 

Which had introduced another bug in which if you had a custom workspace title defined, it would show the default AND the custom one since we never skipped while doing our iterations.

I think iterators should still be on the table for operations which are high frequency working with objects of large size. However, for icon maps such as this instance, the tangible benefit is hard to justify the complexity of the implementation.

I've gone ahead an created a branch called `709-updates` that you can merge into this; which simplifies some of the code and borrows a bit from your implementations. It still uses object copies but as I've stated, since these object copies are ephemeral, they're negligible when it comes to memory usage.

davfsa:
Merged!

PR #713, opened 2025-01-11
Fix matching of scaled monitors

### LINKED ISSUE #636: Bar is Displayed on wrong monitors (Dual Monitor)
**Describe the bug**
Using a dual monitor laptop, the bars are being displayed on the wrong screens. 

The top (main) screen shows the bar for the bottom screen (workspaces shows the bottom screen's workspaces) and vice versa.

**Expected behavior**
Bars should be swapped around to be in the correct screens.

**Screenshots**
image_2024-12-27_083547503

 - Distribution: Arch Linux

Sawy7:
Hi, I made an attempt at fixing broken monitor ID matching when using scaling. This works well on my Nix workstation with one 4k and one FHD monitor.

I'll happily make changes in the implementation if it is needed. Please let me know!

Closes #636


Charging1948:
Tested it. Works perfectly for my setup.

simonkampe:
Sorry, did not meant to reference your PR in my own fork, I tested yours and it works on my machine. I just wanted to run my own fix until this is merged :)

Diaoul:
@Jas-SinghFSU do you think we can get this merged? Assuming it works as well for single monitors setup we should be good :+1:

simonkampe:
I tried it both single and multi monitor, seems to work fine.

Jas-SinghFSU:
@Sawy7 Could you please fix the linter issues when you get a chance :) 
```
npm run lint:fix
```

uwidev:
It seems to not resolve this issue on my end when building from the this pull request.
image

I have different sized monitors, but I changed their scaling such that after scaling, they are at the same resolution. One scales up while the other scales down to try to meet in the middle.
```
monitor=DP-3,2560x1440@144,0x0,1.25,bitdepth,10
monitor=HDMI-A-1,1920x1080,-1152x-896,0.9375,transform,1,bitdepth,10
```

Sawy7:
> @Sawy7 Could you please fix the linter issues when you get a chance :)
> 
> ```
> npm run lint:fix
> ```

Ran the lint command and rebased. Would love to look at @uwidev's issue, but don't have the time right now. If you'd like to postpone this patch, I'd understand.

wistfulbrick:
I may well be wrong in some of my above statements, I've only been looking through this project and GDK for the last few days. If I am wrong please feel free to correct me, the clarifications would benefit us all :)

uwidev:
As per request from @wistfulbrick 
> Could you share what version of GDK is running in hyprpanel --inspector?

image

> Do you use other/your own AGS components that perhaps use a more recent version of AGS (no idea if later AGS versions upgraded to GDK4 or not).

This is the out-of-box experience with no modifications.

wistfulbrick:
> This is the out-of-box experience with no modifications.
> 
> Also... did something get pushed that fixes which bar goes to which monitor? They're appearing on the correct monitor now for me. I DID perform a system since the my last post...

Bizarre, perhaps someone more knowledgeable with GDK might be able to figure it out. The GDK resolution on your Dell monitor makes even less sense now, unless the monitor resolution/scale in hyprland has also changed since.

uwidev:
> > This is the out-of-box experience with no modifications.
> > Also... did something get pushed that fixes which bar goes to which monitor? They're appearing on the correct monitor now for me. I DID perform a system since the my last post...
> 
> Bizarre, perhaps someone more knowledgeable with GDK might be able to figure it out. The GDK geometry on your Dell monitor makes even less sense now, unless the monitor resolution/scale in hyprland has also changed since.

You reminded me, I forgot about it, but since my very first post, I have also changed the scaling of my monitors, so that screenshot is actually inaccurate. My apologies. Here are the actual settings when changing back to the resolution where the issue occurred.

image

wistfulbrick:
@Sawy7 I've tried to reproduce @uwidev 's issue in every way I could with my hardware but can't get my machine to replicate that fractional GDK geometry they were getting.

That's definitely their issue though, their GDK geometry height/width GDK was fractionally scaled but your code (and mine) assumes that it will be scaled with the non-fractional scaleFactor. Your code applies the scaleFactor to the hyprland monitor height/width to match it in the key (and mine unscales the gdk geometry with it). Either way, the two keys won't match. 

I think @uwidev 's issue is some other bug basically, unless someone else can reproduce it. I don't think it should prevent you or @Jas-SinghFSU from considering merging this.

tixwho:
Hello there, luckily enough I also have a Dell monitor involved in my dual-monitor setup, and the build has not fixed the issue (workspace bar still on the wrong monitor). The two monitors are on the same scaling and of the same size, scaling set to 1.5. 

I'm not totally sure if it is something

Here are the logs:

monitor setup in `hyprland.conf`:
```
$MONITOR_L = desc:LG Electronics LG ULTRAGEAR+ 310NTXRBV490 #landscape
$MONITOR_R = desc:Dell Inc. DELL U2723QE 6K155P3 #portrait

monitor = $MONITOR_L, 3840x2160@144, 0x0, 1.5
# position use scaled one, so 3840/1.5=2560
monitor = $MONITOR_R, preferred, 2560x-406, 1.5, transform, 3 
monitor = , preferred, auto, 1
```

`xrandr --listmonitors`

``` 
0: +DP-2 2160/600x3840/340+0+0  DP-2
1: +DP-3 3840/600x2160/340+2160+0  DP-3
```

`hyprctl monitors`

```
Monitor DP-2 (ID 0):
	3840x2160@59.99700 at 2560x-406
	description: Dell Inc. DELL U2723QE 6K155P3
	make: Dell Inc.
	model: DELL U2723QE
	serial: 6K155P3
	active workspace: 9 (9)
	special workspace: 0 ()
	reserved: 0 51 0 0
	scale: 1.50
	transform: 3
	focused: no
	dpmsStatus: 1
	vrr: false
	solitary: 0
	activelyTearing: false
	directScanoutTo: 0
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 3840x2160@60.00Hz 3840x2160@29.98Hz 2560x1440@59.95Hz 2048x1280@60.20Hz 2048x1152@60.00Hz 1920x1200@59.88Hz 2048x1080@24.00Hz 1920x1080@60.00Hz 1920x1080@59.94Hz 1920x1080@50.00Hz 1920x1080@23.98Hz 1600x1200@60.00Hz 1680x1050@59.95Hz 1280x1024@75.03Hz 1280x1024@60.02Hz 1280x800@59.81Hz 1152x864@75.00Hz 1280x720@59.94Hz 1280x720@50.00Hz 1024x768@75.03Hz 1024x768@60.00Hz 800x600@75.00Hz 800x600@60.32Hz 720x576@50.00Hz 720x480@59.94Hz 640x480@75.00Hz 640x480@59.94Hz 640x480@59.93Hz 

Monitor DP-3 (ID 1):
	3840x2160@143.99899 at 0x0
	description: LG Electronics LG ULTRAGEAR+ 310NTXRBV490
	make: LG Electronics
	model: LG ULTRAGEAR+
	serial: 310NTXRBV490
	active workspace: 1 (1)
	special workspace: 0 ()
	reserved: 0 51 0 0
	scale: 1.50
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: true
	solitary: 0
	activelyTearing: false
	directScanoutTo: 0
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 3840x2160@144.00Hz 3840x2160@94.98Hz 3840x2160@60.00Hz 2560x1440@143.97Hz 1920x1080@143.98Hz 1920x1080@60.00Hz 1920x1080@59.94Hz 1280x720@59.94Hz 1024x768@60.00Hz 800x600@60.32Hz 720x480@59.94Hz 640x480@59.95Hz 640x480@59.94Hz 640x480@59.93Hz
```
GTK debug:
1ed83796dc224fa45a6f7fef986345a3

wistfulbrick:
@tixwho Interesting, thank you for replicating it! Again the Geometry height width has also been fractionally scaled, which is what is causing the keys to not match. Your output is actually super useful for narrowing down possible culprits. 

So considering that _both_ monitors in your GDK output are fractionally scaled, and whilst you have a similar dell monitor, uwidev's second monitor was an Ankor and your second monitor is an LG. I think we can rule out anything specific to monitor brands. Additionally, uwidev's issue had both hdmi and DP outputs and you're replicating the issue with two DP monitors, so we can _probably_ rule out handling of different display connector types. 

I think I know what the issue is here, I'm just collecting my evidence to present my theory. Are you and @uwidev running any gtk4 applications? If you're on Arch I'm pretty sure that both astal3 and astal4 and related gdk3/4 libraries get installed with the ags-hyprpanel-git. Can you let me know if you have both gtk-launch and gtk4-launch in your PATH and what versions?

wistfulbrick:
So I'm absolutely _not_ a GTK/GDK expert, I've just been reviewing the source code for GDK3 and GDK4 and how they handle scaling and setting geometry. Please take what I'm about to say with a grain of salt. 

In both versions the monitor scaling and geometry is handled by the apply_monitor_change function:
- In GDK3 this is in [gdkscreen-wayland.c](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/wayland/gdkscreen-wayland.c#L1555)
- In GDK4 this is in [gdkdisplay-wayland.c](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-4-6/gdk/wayland/gdkdisplay-wayland.c#L2314)

How the two versions handle scaling is quite different. 

In GDK3 it scales the gdkmonitor geometry (its position and size) by the scale factor (the wayland fractional scale rounded up to nearest integer). Then the gdkscreen and surfaces within(?) render the gui components within that pre scaled monitor geometry space.  

In GDK4 the gdkmonitor geometry is not scaled at all, it is set to the original resolution height/width of the wayland display. The scaling happens in a function [update_scale](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-4-6/gdk/wayland/gdkdisplay-wayland.c#L2248) which (I think) goes through all the surface/display components and applies the respective scale factor for the monitor that the surface outputs to. 

What this means is that GDK3/4 will set the monitor gdkgeometry values differently: 
- GDK3 geometry = wayland_resolution / scaleFactor
- GDK4 geometry = wayland_resolution

I know that GDK supports having multiple GDK major versions on the same system. In my case I'm on NixOS and it looks like my setup seems tightly limited to GDK3 only, my GDK prefix in inspector is `/nix/store/22krgnw10vrhdfhrfkgr1zp8kwlh1xfv-gtk+3-3.24.43`. But on arch installs I believe both astal3/4 and gdk3/4 are installed (in your case to `/opt`). 

 I think what's happening is that somehow when there are both GDK3 and GDK4 applications running, their GDKMonitor geometry will be in conflict with each other? Or one wins out over the other somehow? So we can run a HyprPanel, a GDK3 application but the gdkmonitor geometry may well be set by a GDK4 application elsewhere on the system.

uwidev:
@wistfulbrick correct, I am on Arch. I don't recall specifically installing any gtk4. Regardless, here's what I tried to gather (wasn't too sure, had to ask LLMs for help).
```sh
> which gtk-launch; gtk-launch --version
/usr/bin/gtk-launch
3.24.48
> which gtk4-launch; gtk4-launch --version
/usr/bin/gtk4-launch
4.16.12
```

Installed packages related to gtk4.
```txt
> pacman -Qs gtk4
local/appmenu-glib-translator-git 24.05.r6.g218bb4a-1
    GLib-based library for translating DBusMenu-exported menus into GMenuModels. Used
    for vala-panel's system tray and all variants of vala-panel-appmenu. Can be used
    with GTK3 or GTK4.
local/gtk4 1:4.16.12-1
    GObject-based multi-platform GUI toolkit
local/gtk4-layer-shell 1.0.4-1
    Library to create panels and other desktop components for Wayland
local/libastal-4-git r736.32348e7-1 (libastal)
    Building blocks for building desktop shells using gtk4.
local/libportal-gtk4 0.9.0-1
    GIO-style async APIs for most Flatpak portals - GTK 4 backend
```
Any processes potentially utilizing gtk4
```sh
> grep -Rl 'libgtk-4' /proc/*/maps 2>/dev/null
/proc/290752/maps
```
Which corresponds to [easyeffects](https://github.com/wwmm/easyeffects).

An ldd for executable linked to libgtk-4.
```
> ldd /usr/bin/* 2>/dev/null | grep libgtk-4
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x00007a00e9800000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x0000767315800000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x0000790e05c00000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x0000711482c00000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x00007413ba600000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x00007fb397600000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x0000784864200000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x000075dbc2600000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x000078349ba00000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x00007d37f5c00000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x00007e1717000000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x00007852d5a00000)
	libgtk-4.so.1 => /usr/lib/libgtk-4.so.1 (0x0000701f9b600000)
```

I tested without easyeffects runing, no changes.

Not too sure if this is relevant or helpful, but with hyprpanel open, changing the resolution configs does not result in any changes to where the bars are placed.

Here are is my working resolution configs
```
# Scale down 1440p to 1080p, resulting in lower DPI
monitor=DP-3,2560x1440@144,0x0,1.3333,bitdepth,10
monitor=HDMI-A-1,1920x1080,-1080x-840,1,transform,1,bitdepth,10
```

And not working resolution configs
```
# Scale up and down to meet in the middle, moderate DPI
monitor=DP-3,2560x1440@144,0x0,1.25,bitdepth,10
monitor=HDMI-A-1,1920x1080,-1152x-896,0.9375,transform,1,bitdepth,10
```

wistfulbrick:
@uwidev nice digging! Though I'm not 100% sure whether uninstalling all GTK4 apps would prevent the behavior if you still have the libraries (I still don't understand why this behavior is happening). 

>Not too sure if this is relevant or helpful, but with hyprpanel open, changing the resolution configs does not result in any changes to where the bars are placed.

Yeah I found this as well with my testing, you have to restart the hyprland session for it to take effect (in my case I just did a full reboot after changing hyprland resolution). Could you set your non-working resolution configs and restart hyprland or reboot and test it again with easyeffects removed? It would be good to eliminate it as a possibility.

uwidev:
@wistfulbrick 
image

Easyeffects uninstalled `pacman -Rns easyeffects`.

wistfulbrick:
@uwidev Thank you, it doesn't seem to have effected it. I'm also realizing I was looking at an older GDK4 version :sweat: a lot of what I previously said on GDK4 regarding geometry doesn't apply, sorry! 

I do still think this issue has something along these lines.. I'm probably way out of my depth here but I'll keep reading through the codebases and see if I can figure it out.

ppenguin:
For me this branch fixes the issue too (in my case two identical monitors scaled at `1.5`)

wistfulbrick:
Aha! I think this is probably the lead we're looking for on this https://gitlab.gnome.org/GNOME/gtk/-/issues/3066
```
The problem is that gdk_monitor_get_geometry() will return different results depending whether the Wayland compositor scales the framebuffer (e.g. for fractional scaling) or not.
```

`gdk_monitor_get_geometry() ` is what returns the gdkmonitor geometry both in this code, but also the inspector windows. 

As quoted from: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/master/unstable/xdg-output/xdg-output-unstable-v1.xml#L115
```
	For example, for a wl_output mode 3840×2160 and a scale factor 2:

	- A compositor not scaling the surface buffers will advertise a
	  logical size of 3840×2160,

	- A compositor automatically scaling the surface buffers will
	  advertise a logical size of 1920×1080,

	- A compositor using a fractional scale of 1.5 will advertise a
	  logical size to 2560×1620.
```

GDK will only scale the logical geometry size by the gdk scaleFactor if it finds the geometry for xdg_output and wl_output aren't the same [gdk4-ref](https://gitlab.gnome.org/GNOME/gtk/-/blob/4.17.4/gdk/wayland/gdkdisplay-wayland.c#L2410).

This is a far more likely explanation than my misunderstanding of GDK3/4 :grin: 

The Arch docs have some info around this https://wiki.archlinux.org/title/HiDPI I'm curious whether you folks have this set up:
```
gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"
```

wistfulbrick:
Another strange thing I found was that if I `nix-shell -p gtk4` and then run `gtk4-demo-application` I can run a GTK4 inspector session by pressing CTRL+SHIFT+D. 

In this GTK4 inspector window, my gdkmonitor geometries _are_ fractionally scaled correctly (and the positions are correct too). So this _might_ just be an issue with GDK3?

tixwho:
@wistfulbrick Thanks for the follow up! No, I only have KDE and hyprland installed so I never checked the gnome section.

Running the `gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"` returned `No such schema “org.gnome.mutter”`. Rightfully so , because mutter wasn't installed at all! 
So I attempted to use the permanent dconf configuration method described in Archwiki HiDPI [2.1.1.1](https://wiki.archlinux.org/title/HiDPI#Wayland), that didn't work. A mutter installation after reboot wasn't helpful as well.

A follow up here, here is a side-by-side inspector comparison between a GTK-4 application of mine (trayscale) and GTK-3 based hyprpanel. The GTK4 application is on the left side.

2025-02-09-154901_hyprshot

ppenguin:
I fail to understand why scaling should be relevant to the monitor selection algorithm anyway as seems to be the case, but I admit I haven't looked into any details at all... 

I haven't done any additional `gsettings` stuff, just some theming with `home-manager` and `stylix`.

As an additional data point: GTK inspector on the demo app seems to identify the same as `hyprctl`, so I don't understand why in such case the unpatched `hyprpanel` swapped the workspaces in the first place?

image

image

wistfulbrick:
> I fail to understand why scaling should be relevant to the monitor selection algorithm anyway as seems to be the case, but I admit I haven't looked into any details at all...

It's not a hyprland workspace issue, it's an issue with how GDK tracks monitor ids, to know which bars (and thus which workspace numbers) to render to the correct monitor. 

The context you're looking for is [here](https://github.com/Jas-SinghFSU/HyprPanel/blob/0d5f80ff5cd525b8f27adfb84cef67d90e3d7f10/src/components/bar/utils/monitors.ts#L73). 

Hyprpanel has to a decent amount of work to attempt to pair gdk monitor ids with hyprland monitor ids. It matches them by a key that's based on monitor model, height/width and scale. Unfortunately the gdk height/width is proving to be equally as slippery!

wistfulbrick:
Thank you folks for testing out the mutter setting, I really am just throwing things out here at the moment. I think there's _some_ setting/config option either in hyprland or gdk that's being applied (or not) that's causing the variation in behavior. As soon as you have found a setting that changes whether the GDK3 inspector geometry scales with fractional scale or scale factor, please post here. 

There's two things we could possibly try code wise, we could try and anticipate and match keys against both a fractionally scaled geometry and a scale factor scaled geometry. Or we could try updating Hyprpanel to use astal4/gdk4 (which may break a few other things) as it _seems_ GDK4 is more reliable in fetching accurate monitor information.

wistfulbrick:
Hahaha, I just spent the afternoon trying to hack HyprPanel to work with gdk4 to no success. Really should have expected tbh, it's a huge amount of work and I'm not remotely qualified to do it!

There's a ton of changes to make, some of the old GTK3 widgets HyprPanel used like Gtk.Menu have been deprecated and require work to convert to alternatives, it looks like you have to declare external GTK4 widgets in a new/different way, there's a good few gdk functions that have been deprecated needing sometimes additional logic or testing. It's not a complete rewrite but it's _substantial_. Probably not a fair ask on the maintainers!

wistfulbrick:
Instead I think we can a workaround for this behavior that tries to match against both outcomes:
- Where the GDK key height/width are scaled by the scaleFactor
- Where the GDK key height/width are scaled by the scale (fractional) 

I've tested this on my local flake of HyprPanel and it works for me, which isn't conclusive of course. Obviously the real test will be whether it works for others with fractional geometry: 
```
gdk key setup same as master
...
// First pass: Strict matching including the monitor index (i.e., hypMon.id === monitor + resolution+scale criteria)
    const directMatch = hyprlandService.get_monitors().find((hypMon) => {
        const isVertical = hypMon?.transform !== undefined ? hypMon.transform % 2 !== 0 : false;

        const gdkScaleFactor = Math.ceil(hypMon.scale);

        const width = isVertical ? hypMon.height : hypMon.width;
        const height = isVertical ? hypMon.width : hypMon.height;

        const scaleFactorWidth = Math.trunc(width / gdkScaleFactor);
        const scaleFactorHeight = Math.trunc(height / gdkScaleFactor);
        const scaleFactorKey = `${hypMon.model}_${scaleFactorWidth}x${scaleFactorHeight}_${gdkScaleFactor}`;

        const scaleWidth = Math.trunc(width / hypMon.scale);
        const scaleHeight = Math.trunc(height / hypMon.scale);
        const scaleKey = `${hypMon.model}_${scaleWidth}x${scaleHeight}_${gdkScaleFactor}`;

        const keyMatch = (gdkMonitor.key === scaleFactorKey || gdkMonitor.key === scaleKey);

        return keyMatch && !usedHyprlandMonitors.has(hypMon.id) && hypMon.id === monitor;
    });

... and similar for the second pass
```

Only issue is we're now matching against two keys so there's _maybe_ the potential for more key conflicts, currently doing the mental gymnastics to figure out in what additional circumstances this would occur.

PR #720, opened 2025-01-12
Ensure notification service is available when sending a notification

davfsa:
This is a bit of a weird patch for a fundamental order-of-execution issue.

The bundler (esbuild) can sometimes place the `notifdService` instantiation bellow the code calling `Notify`, which would causes the `notify-send` execution to fail. To avoid this, we can make it seem like `notifdService` is being used inside `Notify`, so that it is always placed above the usage of the function, ensuring that we always have a notification daemon running

---

This issue has become apparent to me as I have upgraded Hyprpanel, as sometimes the `"swww is a missing dependency"` notification would sometimes show up on startup and sometimes not (it would depend on where the bundler would place the piece of code).

Additionally, this will allow #684 to send notifications when creating the configuration, for the same reason as listed above.

---

If anybody knows a better way to indicate to the bundler to always place `src/lib/constants/services.ts` at the top of the bundle, then that would be a way better fix, but I have been unable to find a way to do it


src/lib/utils.ts
```diff
@@ -11,6 +11,7 @@ import options from '../options';
 import { Astal, Gdk, Gtk } from 'astal/gtk3';
 import AstalApps from 'gi://AstalApps?version=0.1';
 import { exec, execAsync } from 'astal/process';
+import { notifdService } from './constants/services.ts';
```
Jas-SinghFSU, line 14:
```suggestion
import AstalNotifd from 'gi://AstalNotifd?version=0.1';

const notifdService = AstalNotifd.get_default();
```

PR #732, opened 2025-01-17
feat(hypridle): fixes #643 by making commands customizable

### LINKED ISSUE #643: Hypridle module not working
**Describe the bug**
The `hypridle` module switches from the `disabled` state to the `enabled` state with no manual interaction. 

**To Reproduce**

1. Click on `hypridle` module to switch from `enabled` to `disabled` (in order to disable screen locking).
2. Continue to interact with machine as usual
3. In roughly 1 minute, the state and icon will switch back to the `enabled` state.

**Expected behavior**
I expect that `hypridle` will be toggled correctly and the bar module correctly reflects the current activation state

**Screenshots**
N/A, no relevant console output was seen when triggering bug from `hyprpanel` launched from ternminal

 - Distribution: NixOS 24.11

**Additional context**

Here's my `hypridle` [config](https://github.com/GideonWolfe/nix/blob/main/configs/users/gideon/configs/hypridle.nix)

```nix
{ config, lib, pkgs, ... }: {
  # Enable hypridle
  services.hypridle = {
    enable = true;
    settings = {
      listener = [
        # warning before screen lock
        {
          timeout = 240;
          on-timeout =
            "notify-send -u critical 'Screen Will Lock' 'One minute remaining'";
        }
        # will lock screen
        {
          timeout = 300;
          on-timeout = "hyprlock";
        }
        # will enter hibernation
        {
          timeout = 600;
          on-timeout = "hyprctl dispatch dpms off";
          on-resume = "hyprctl dispatch dpms on";
        }
      ];
    };
  };
}

```

FerranAD:
Hypridle can be started both by running the binary at startup `exec-once = hypridle` or a user service with systemd `systemctl --user enable --now hypridle.service` as noted on the documentation [usage section](https://github.com/hyprwm/hypridle?tab=readme-ov-file#usage). Hyprpanel custom module for Hypridle fails when it is started with system ad reported on https://github.com/Jas-SinghFSU/HyprPanel/issues/643, which makes sense when looking at the commands used to start, stop and check status of Hypridle. See
https://github.com/Jas-SinghFSU/HyprPanel/blob/88609f7e4c244326face34cb992a79053f2c7810/src/components/bar/modules/hypridle/helpers/index.ts#L9 and https://github.com/Jas-SinghFSU/HyprPanel/blob/88609f7e4c244326face34cb992a79053f2c7810/src/components/bar/modules/hypridle/helpers/index.ts#L40
Running `pkill hypridle` when it has been started with systemd will only temporarily stop Hypridle and systemd is gonna restart it in a few seconds, resulting in the reported behavior at the above mentioned issue.

This pull request adds the possibility to customize said commands but maintains the default values they had for the sake of backwards compatibility. Except for NixOS where it does set commands using systemd since it seems to be the most common and recommended way.

I have tested it still works with the default values as it did until now when not using systemd. 

For NixOS users that start Hypridle with Hyrpland's `exec-once` and  want to keep the old behavior, set the options to the following:

```nix
bar.customModules.hypridle.startCommand = mkStrOption "nohup hypridle > /dev/null 2>&1 &";
bar.customModules.hypridle.stopCommand = mkStrOption "pkill hypridle";
bar.customModules.hypridle.isActiveCommand = mkStrOption "pgrep -x 'hypridle' &>/dev/null && echo 'yes' || echo 'no'";
```

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -203,6 +203,9 @@ in
       bar.customModules.hypridle.rightClick = mkStrOption "";
       bar.customModules.hypridle.scrollDown = mkStrOption "";
       bar.customModules.hypridle.scrollUp = mkStrOption "";
+      bar.customModules.hypridle.startCommand = "systemctl --user start hypridle.service";
+      bar.customModules.hypridle.stopCommand = "systemctl --user stop hypridle.service";
+      bar.customModules.hypridle.isActiveCommand = "systemctl --user status hypridle.service | grep -q 'Active: active (running)' && echo 'yes' || echo 'no'";
       bar.customModules.hyprsunset.label = mkBoolOption true;
       bar.customModules.hyprsunset.middleClick = mkStrOption "";
       bar.customModules.hyprsunset.offIcon = mkStrOption "󰛨";
--- a/src/components/bar/modules/hypridle/helpers/index.ts
+++ b/src/components/bar/modules/hypridle/helpers/index.ts
@@ -1,12 +1,15 @@
+import options from 'src/options';
 import { execAsync, Variable } from 'astal';
+
+const { startCommand, stopCommand, isActiveCommand } = options.bar.customModules.hypridle;
 
 /**
  * Checks if the hypridle process is active.
  *
- * This command checks if the hypridle process is currently running by using the `pgrep` command.
+ * This command checks if the hypridle process is currently running by using the configured `isActiveCommand`.
  * It returns 'yes' if the process is found and 'no' otherwise.
  */
-export const isActiveCommand = `bash -c "pgrep -x 'hypridle' &>/dev/null && echo 'yes' || echo 'no'"`;
+export const isActiveCmd = `bash -c "${isActiveCommand.get()}"`;
 
 /**
  * A variable to track the active state of the hypridle process.
@@ -21,8 +24,8 @@ export const isActive = Variable(false);
  * @param isActive A Variable<boolean> that tracks the active state of the hypridle process.
  */
 const updateIsActive = (isActive: Variable<boolean>): void => {
-    execAsync(isActiveCommand).then((res) => {
-        isActive.set(res === 'yes');
+    execAsync(isActiveCmd).then((res) => {
+        isActive.set(res.trim() === 'yes');
     });
 };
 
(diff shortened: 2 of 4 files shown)


GideonWolfe:
Just wanted to add that this might also be a path that could be taken with the Power Profile Management module? I'm seeing a similar thing where I can select a profile but it pretty much immediately switches back to "balanced".

nix/module.nix
```diff
@@ -203,6 +203,9 @@ in
       bar.customModules.hypridle.rightClick = mkStrOption "";
       bar.customModules.hypridle.scrollDown = mkStrOption "";
       bar.customModules.hypridle.scrollUp = mkStrOption "";
+      bar.customModules.hypridle.startCommand = mkStrOption "nohup hypridle > /dev/null 2>&1 &";
+      bar.customModules.hypridle.stopCommand = mkStrOption "pkill hypridle";
+      bar.customModules.hypridle.isActiveCommand = mkStrOption "pgrep -x 'hypridle' &>/dev/null && echo 'yes' || echo 'no'";
```
PerchunPak, line 208: It is safe to assume that hypridle is managed by `programs.hypridle.enable` (== systemd) on NixOS
```suggestion
bar.customModules.hypridle.startCommand = "systemctl --user start hypridle.service";
bar.customModules.hypridle.stopCommand = "systemctl --user stop hypridle.service";
bar.customModules.hypridle.isActiveCommand = "systemctl --user status hypridle.service | grep -q 'Active: active (running)' && echo 'yes' || echo 'no'";
```
```

PerchunPak: That's what I am talking about, you should default to systemd commands on NixOS

FerranAD: Yeah that makes sense. Changed!

PerchunPak: looks like you forgot to commit this

PerchunPak (approved):
Diff LGTM, rebuilding my hyprpanel with this patch

FerranAD:
@GideonWolfe looks good to you aswell?

nix/module.nix
PerchunPak, line 208: Looks like you forgot to commit my suggestion, even though you agreed to it

Jas-SinghFSU:
I'm wondering if we're better off using the idle inhibitor provided by Astal rather than Hypridle.

PerchunPak:
> I'm wondering if we're better off using the idle inhibitor provided by Astal rather than Hypridle.

I agree, but let's merge this one for now

nix/module.nix
FerranAD: Oh you are right, I edited the PR description but forgot to commit the changes. Should be changed now!

Jas-SinghFSU:
I've decided to use the Astal inhibitor module so we don't have to manage the Hypridle process manually now. Keeps things super simple since we can toggle the inhibitor without having to kill the Hypridle process.
#776

PR #782, opened 2025-02-19
Add "workspaces per monitor" setting

dotaxis:
When using zero-based workspaces, workspaces are incorrectly mapped if there are empty workspaces before active ones.

For example, in this screenshot I am actually on workspace 6, but since workspaces 2-5 are empty, HyprPanel shows that I'm on workspace 2.
image

This PR adds an option to specify how many workspaces should be on each monitor to remedy this. It defaults to 10.
image

I have only tested this on 2 monitors because that's all I have. The math may need to be tweaked to support more monitors.


src/components/bar/modules/workspaces/helpers/utils.ts
```diff
@@ -271,5 +276,10 @@ export const renderLabel = (
         return getWsIcon(wsIconMap, i);
     }
 
-    return workspaceMask ? `${index + 1}` : `${i}`;
+    if (workspaceMask) {
+        const j = workspacesPerMonitor.get();
```
Jas-SinghFSU, line 280:
```suggestion
        const j = zeroBasedModulo.get();
```

src/components/bar/modules/workspaces/helpers/utils.ts
```diff
@@ -8,7 +8,12 @@ const hyprlandService = AstalHyprland.get_default();
 const { monochrome, background } = options.theme.bar.buttons;
 const { background: wsBackground, active } = options.theme.bar.buttons.workspaces;
 
-const { showWsIcons, showAllActive, numbered_active_indicator: wsActiveIndicator } = options.bar.workspaces;
+const {
+    showWsIcons,
+    showAllActive,
+    numbered_active_indicator: wsActiveIndicator,
+    workspacesPerMonitor,
```
Jas-SinghFSU, line 15:
```suggestion
    zeroBasedModulo,
```

dotaxis:
This should now work when there are a different numbers of of static workspaces used per monitor. Could use some additional testing on other configurations.

Jas-SinghFSU:
Closing in favor of your effort that will implement mapping workspace numbers instead.

Issue #366, opened 2024-10-24
FEATURE: Box Shadow and screen corners


rubiin:
**Is your feature request related to a problem? Please describe.**

Add box shadow which can be configured
Also screen corners widget (https://github.com/Aylur/dotfiles/blob/main/ags/widget/bar/ScreenCorners.ts)
Also box shadow

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #784, opened 2025-02-19
Add nix options for shadows

leszkolukasz:
In #366 shadows were added but without nix options to configure them.

This PR adds missing options. Documentation also needs to be updated to reflect new options.


Jas-SinghFSU:
Ty :heart:

PR #785, opened 2025-02-19
Add charging state notifications

Dirtyjimmy:


src/lib/behaviors/batteryWarning.ts
```diff
@@ -8,14 +9,28 @@ export function warnOnLowBattery(): void {
     let sentLowNotification = false;
     let sentHalfLowNotification = false;
 
+    // Charging state notifications
     batteryService.connect('notify::charging', () => {
-        // Reset it when the battery is put to charge
         if (batteryService.charging) {
-            sentLowNotification = false;
-            sentHalfLowNotification = false;
```
Jas-SinghFSU, line 15: Removing these lines will break the low battery notifications.

Jas-SinghFSU:
I'm trying to understand the need for notifications since the battery module indicates the battery charging status. Perhaps this is best kept as a script that does a `notify-send`.

PR #789, opened 2025-02-21
Add 'donottouch' autohide option

atikiNBTW:
Adds small "do not touch" autohide option that allows the panel to be toggled without always reappearing.

Maybe change the name idk, feel free to suggest one


atikiNBTW:
Maybe I'm doing something wrong, but I couldn't setup hiding and showing the bar with a shortcut, because with every currently existing autohide mode it randomly showed up when a new window was opened.

Jas-SinghFSU:
Ah, I see. Yea that was a bug and it has been fixed. If you update, we can close this PR if it is no longer an issue for you.

atikiNBTW:
yeah, the issue seems to be gone now, thanks

PR #791, opened 2025-02-23
Add region screen recording

dotaxis:
Switch to wf-recorder and add region recording with slurp.

AUR package will need to be updated to replace `gpu-screen-recorder` with `wf-recorder`.

A flatpak is not available for wf-recorder but I don't think that's a big deal. It looks like only Fedora relied on the gpu-screen-recorder flatpak but wf-recorder is available via `dnf`.

image

Works on my machine™️

Needs testing on different hardware configurations, specifically Nvidia.


scripts/screen_record.sh
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
+# Requires wf-recorder: https://github.com/ammen99/wf-recorder
 
 outputDir="$HOME/Videos/Screencasts"
+defaultSink=$(pactl info | grep "Default Sink" | awk -F ': ' '{print $NF}')
+WF_RECORDER_OPTS="--audio $defaultSink --codec av1_vaapi"
```
Jas-SinghFSU, line 6:
```suggestion
defaultSink=$(pactl get-default-sink)
WF_RECORDER_OPTS="--audio=$defaultSink.monitor -c libx264rgb"
```
- Changed the default audio input to be the desktop audio of the default sink.
- Changed the video codec to software instead of hardware accelerated for now as it works more out of the box. We can make this more configurable in a separate PR

scripts/screen_record.sh
```diff
@@ -18,24 +21,26 @@ startRecording() {
 
     target="$2"
 
-    outputFile="recording_$(date +%Y-%m-%d_%H-%M-%S).mp4"
-    outputPath="$outputDir/$outputFile"
+    outputFile="recording_$(date +%Y-%m-%d_%H-%M-%S)"
+    outputPath="$outputDir/${outputFile}.mp4"
+    logFile="$outputDir/${outputFile}.log"
```
Jas-SinghFSU, line 26:
```suggestion
```

scripts/screen_record.sh
```diff
@@ -18,24 +21,26 @@ startRecording() {
...
     mkdir -p "$outputDir"
 
-    if [ -z "$target" ]; then
-        echo "Usage: $0 start screen [screen_name]"
+    if [ "$target" == "screen" ]; then
+        monitor_info=$(hyprctl -j monitors | jq -r ".[] | select(.name == \"$3\")")
+        echo "Monitor Selected: $3" > "$logFile"
```
Jas-SinghFSU, line 31:
```suggestion
```

scripts/screen_record.sh
```diff
@@ -18,24 +21,26 @@ startRecording() {
...
+        w=$(echo $monitor_info | jq -r '.width')
+        h=$(echo $monitor_info | jq -r '.height')
+        x=$(echo $monitor_info | jq -r '.x')
+        y=$(echo $monitor_info | jq -r '.y')
+        wf-recorder $WF_RECORDER_OPTS --geometry "${x},${y} ${w}x${h}" --file "$outputPath" 1>>"$logFile" 2>&1 &
```
Jas-SinghFSU, line 36:
```suggestion
        wf-recorder $WF_RECORDER_OPTS --geometry "${x},${y} ${w}x${h}" --file "$outputPath" &
```

scripts/screen_record.sh
```diff
@@ -18,24 +21,26 @@ startRecording() {
...
+    elif [ "$target" == "region" ]; then
+        wf-recorder $WF_RECORDER_OPTS --geometry "$(slurp)" --file "$outputPath" 1>"$logFile" 2>&1 &
```
Jas-SinghFSU, line 38:
```suggestion
        wf-recorder $WF_RECORDER_OPTS --geometry "$(slurp)" --file "$outputPath" &
```

dotaxis:
I thought it was pretty lame that I had to use `awk` to get the default sink from `pactl`. I guess I didn't.

Should be good to merge now but the AUR package will need to be updated. I'll open a separate PR to allow configuring codecs and sinks once we decide the best way to handle that.

PR #795, opened 2025-02-25
feat: add world clock

make-42:
Implements what is wanted in https://github.com/Jas-SinghFSU/HyprPanel/pull/690 fully.

Theming files might have to be regenerated, I do not know the procedure to do that.


noahfraiture:
Hello, it as been some time. Is there something missing ? I would really like that

PR #809, opened 2025-02-28
fix: scaled width and height on recording

aaronhallaert:
# Issue
when 2 monitors are connected and one of them is scaled (e.g. 1.5), recording button doesn't do anything

```bash
Failed to detect output based on geometry (is your geometry overlapping outputs?)
```

This would work with `slurp -o -d`

# Solution

make sure geometry is identical to `slurp` implementation by extracting scale from `hyprctl monitors`

* depends on `bc`

### DIFF
--- a/scripts/screen_record.sh
+++ b/scripts/screen_record.sh
@@ -29,9 +29,12 @@ startRecording() {
         monitor_info=$(hyprctl -j monitors | jq -r ".[] | select(.name == \"$3\")")
         w=$(echo $monitor_info | jq -r '.width')
         h=$(echo $monitor_info | jq -r '.height')
+        scale=$(echo $monitor_info | jq -r '.scale')
+        scaled_width=$(echo "${w} / ${scale}" | bc)
+        scaled_height=$(echo "${h} / ${scale}" | bc)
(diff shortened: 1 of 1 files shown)


dotaxis:
This broke screen recording because `bc` expects an integer and `hyprctl` reports my scale as `1.00`.

```
╰─λ ./screen_record.sh start screen "DP-2"
./screen_record.sh: line 33: bc: command not found
./screen_record.sh: line 34: bc: command not found
1.00
x
Recording started. Output will be saved to /home/dot/Videos/Screencasts/recording_2025-03-05_23-03-37.mp4

…in repo: HyprPanel/scripts on  workspace-filters [$!⇡2] via  v22.14.0 via  v3.13.2 as 🧙 took 0s
 ╰─λ Bad geometry: 2560,0 x, capturing whole output instead.
Please select an output from the list to capture (enter output no.):
1. Name: DP-1 Description: Ancor Communications Inc ROG PG279Q #ASMHn0YaJQXd (DP-1)
2. Name: DP-2 Description: Dell Inc. DELL G2724D 1SGF4V3 (DP-2)
Enter output no.:Failed to select output, exiting
```

PR #811, opened 2025-03-01
Updated  Cava module to use media module info

EWBennett:
Updated  Cava module to use media icon, media label as tooltip, and d…


Jas-SinghFSU:
Hey @EWBennett, thanks for this PR and for the time you put into it.

Thanks again! <3

Issue #760, opened 2025-02-05
[Feature Request] User defined formatting for workspace display


au-summer:
**Is your feature request related to a problem? Please describe.**
Workspaces module can only display either workspace number or application icon, but not both, which makes it very hard to navigate to the desired workspace.

**Describe the solution you'd like**
Letting user define formatting of workspace display like in waybar. For example `format": "[{name} {windows} ]` will display both workspace number and applications running in the workspace.

![Image](https://github.com/user-attachments/assets/e3557b11-567d-4e07-b7f2-163d3271ffaf)

SirMaxwellSmart:
I would also appreciate something like this. 👍️

I am currently using the "Map Workspaces to Application Icons" feature which is really neat, however I have a few workspaces dedicated to one specific purpose that I want only show a static icon for whether it has a window in it or not.
With the current setup if you have it map the workspace to the app icons it ignores "Workspace Icon & Color Mappings".

PR #818, opened 2025-03-05
Add custom workspace label formats

dotaxis:
Looking to address https://github.com/Jas-SinghFSU/HyprPanel/issues/760.

Curious if this is moving in the right direction or if you had a different approach in mind.

Still needs to display icons in the proper font via `renderClassnames` and better configuration handling.


dotaxis:
image
Getting closer. I almost made this way more complicated than it needed to be.

dotaxis:
https://github.com/user-attachments/assets/99bfa272-7658-43b6-b2bf-43e616361a2d

Removed some now-redundant options which makes things a bit more clear.

Jas-SinghFSU:
Yeah, this looks good to me. We can consolidate everything into one place which is `customFormat` which I like. We're gonna need to thing about a deprecation strategy so we don't break existing user configs once this is implemented.

Jas-SinghFSU:
Hey Dot, closing PRs for HyprPanel as we've moved over to Wayle.

PR #828, opened 2025-03-12
Refactored RecordingButton to fetch the latest recording path dynamically.

### LINKED ISSUE #826: Screen recording location is hardcoded in the script and changes every update.
**Describe the bug**
in the screen_record.sh script the outputDir is hardcoded to $HOME/Videos/Screencasts". i keep my videos in ~/vids. This is a bit annoying and every update i have to edit this script to change the location.

**To Reproduce**
record screen

file saves to "$HOME/Videos/Screencasts"

**Expected behavior**
configurable save location.

**Screenshots**

 - Distribution: [e.g. Arch Linux]
 - Window Manager/Desktop Environment: [e.g. Hyprland]

**Additional context**

siddharthjain25:
Implemented a file picker using Zenity to allow users to choose the save location for their screen recordings after stopping.

Replaced the hardcoded save path with dynamic user input.

Improved the notification system to inform users when recordings are saved or discarded.

Closes #826


scripts/screen_record.sh
```diff
@@ -54,15 +49,17 @@ stopRecording() {
...
 
-    notify-send "Recording stopped" "Your recording has been saved." \
-        -i video-x-generic \
-        -a "Screen Recorder" \
-        -t 10000 \
-        -u normal \
-        --action="scriptAction:-xdg-open $outputDir=Directory" \
-        --action="scriptAction:-xdg-open $recentFile=Play"
+    # Let the user pick a directory using Zenity
+    saveDir=$(zenity --file-selection --directory --title="Select Directory to Save Recording")
```
davfsa, line 55: Instead of forcing zenity onto the user, we can just use whatever files application they want by pulling it from `xdg-desktop-portal-termfilechooser`

Tbh, I do not know quite well how this works just yet, so will be worth some investigation.

scripts/screen_record.sh
```diff
@@ -54,15 +49,17 @@ stopRecording() {
...
+    
+    if [ -n "$saveDir" ]; then
+        mv "$recentFile" "$saveDir/"
+        notify-send "Recording saved" "File saved to $saveDir."
```
davfsa, line 59: Looking at the code this replaces, its missing the action button to open the file location where it's saved, which I think is a nice addition to keep

scripts/screen_record.sh
```diff
@@ -54,15 +49,17 @@ stopRecording() {
...
+    else
+        notify-send "Recording discarded" "File was not saved."
```
davfsa, line 61: Most tools do not notify when the file was not saved, so I think we should probably do the same

scripts/screen_record.sh
```diff
@@ -54,15 +49,17 @@ stopRecording() {
 
     pkill -SIGINT -f wf-recorder
 
-    recentFile=$(ls -t "$outputDir"/recording_*.mp4 | head -n 1)
+    recentFile=$(ls -t "$tempDir"/*.mp4 | head -n 1)
```
davfsa, line 52: Let's make the search a bit unique, as we don't want to just pickup any .mp4 file there is.

We could just always use the same filename and prevent two recordings from happening at the same time

scripts/screen_record.sh
siddharthjain25: I initially tried using xdg-desktop-portal since it’s supposed to work natively with Hyprland and other Wayland environments, but I run into a few issues(plus I tried it on few others as well):

Inconsistent behavior — Even with the right packages (xdg-desktop-portal, xdg-desktop-portal-hyprland), the file picker doesn’t always show up. It defaults to GTK in some cases but still doesn’t trigger the directory selection.

Activation errors — Sometimes we get this error:

"The name is not activatable"
which seems tied to how the portal services communicate, but even after checking the configs, it’s hit-or-miss.

User setup dependency — It relies a lot on the user’s portal setup, so if something’s off in their environment, the picker just won’t work — which feels like too much hassle for something that should be simple.

Because of all this, I think using Zenity might be a better option. It’s lightweight, works without extra setup, and gives a clean GUI for picking directories. Basically, it "just works" without users needing to debug their system.

davfsa: As nice as Zenity might be, its still better to try to get `xdg-desktop-portal` to work as, even if it requires extra setup on peoples part if they didnt set it up correctly (which if people dont have setup by default in their distro, its something they are 99% willing to do, or is already setup by default in all major distributions), we will 100% get people asking for the feature.

I will try to work with you to get this done, dont want to put too much burden on you for it.

If you already toyed around with the idea a bit, I assume that you probably better  at this than me, hahaha. If you can push (or explain) a basic setup for it, I can give it a shot on my system and provide a working patch :)

siddharthjain25: Thanks for the suggestion! You're absolutely right -- using xdg-desktop-portal makes more sense for broader compatibility across distros. I’ll start looking into integrating it for the file picker and notifications. If I run into any issues, I'll definitely reach out for your input. Appreciate your support!

dotaxis:
I am against this implementation.

A config option of where to save them would be much better than relying on a Zenity window for every single recording.

siddharthjain25:
> I don't like this implementation.
> 
> IMO, a config option of where to save them would be much better than relying on a Zenity window for every single recording.

Appreciate the feedback! That makes sense — I’ll add a config option for the save path when I get some free time.

scripts/screen_record.sh
siddharthjain25: > Additionally, as pointed out by @dotaxis, providing a configuration value for default setting the default save location might be a better solution :)

Done as requested.

scripts/screen_record.sh
```diff
@@ -20,49 +17,84 @@ startRecording() {
     fi
 
     target="$2"
+    outputDir=$(eval echo "$3")
```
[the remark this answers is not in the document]
siddharthjain25: I used eval echo "$3" to ensure ~ is expanded to $HOME when users provide a path like ~/Videos. However, I realize eval isn't necessary here and can be avoided for better security.

I can update it to:
outputDir="${3/#\~/$HOME}"

This safely expands ~ while avoiding the risks of eval

scripts/screen_record.sh
```diff
@@ -20,49 +17,84 @@ startRecording() {
...
 
 stopRecording() {
     if ! checkRecording; then
-        echo "No recording is in progress."
+        echo "No recording in progress."
         exit 1
     fi
 
     pkill -SIGINT -f wf-recorder
+    sleep 1
```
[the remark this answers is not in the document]
siddharthjain25: Good point! The sleep 1 is there to ensure wf-recorder has enough time to gracefully terminate before we proceed. Without it, there might be a race condition where the file is still being written when we try to access it immediately after stopping.

I'll add a comment above it to clarify this. Thanks for the suggestion!

src/options.ts
```diff
@@ -1280,6 +1280,9 @@ const options = mkOptions(CONFIG, {
                 interval: opt(2000),
                 enable_gpu: opt(false),
             },
+            recording: {
+                path: opt('$HOME/Videos/Screencasts')
+            },
```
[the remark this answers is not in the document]
siddharthjain25: > New options need to be added to the nix flake too :)
> 
> Also, I don't think we expand `$HOME`, but I do know we expand `~`

I don't use Nix, so I'm not familiar with how to update the flake. Could you guide me on what needs to be added?

scripts/screen_record.sh
```diff
@@ -20,49 +17,84 @@ startRecording() {
...
+        scaled_height=$(awk "BEGIN { print $h / $scale }")
+        x=$(echo "$monitor_info" | jq -r '.x')
+        y=$(echo "$monitor_info" | jq -r '.y')
+
         wf-recorder $WF_RECORDER_OPTS --geometry "${x},${y} ${scaled_width}x${scaled_height}" --file "$outputPath" &
     elif [ "$target" == "region" ]; then
         wf-recorder $WF_RECORDER_OPTS --geometry "$(slurp)" --file "$outputPath" &
     else
-        echo "Usage: $0 start {region|screen [screen_name]}"
+        echo "Usage: $0 start {screen [output_directory] [monitor_name]|region [output_directory]}"
```
dotaxis, line 64: This is a little weird. May want to rearrange our args.

`$0 start {screen [monitor_name]|region} [output_directory]"

scripts/screen_record.sh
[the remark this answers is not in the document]
dotaxis: May want to have a default path here in case $3 is not passed or output_directory is not set.

scripts/screen_record.sh
siddharthjain25: Done as requested.

scripts/screen_record.sh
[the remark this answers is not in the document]
siddharthjain25: I've set a default save location ($HOME/Videos) if no directory is specified and added validation to prevent issues with non-existing directories.

scripts/screen_record.sh
```diff
@@ -21,66 +21,113 @@ startRecording() {
...
+        ;;
+    status)
+        if checkRecording; then
+            echo "recording"
+        else
+            echo "not recording"
+        fi
+        ;;
+    *)
+        echo "Usage: $0 {start [screen monitor_name output_directory | region output_directory] | stop | status}"
```
dotaxis, line 130:
```suggestion
    echo "Usage: $0 {start [screen <monitor_name> | region] <output_directory> | stop | status}"
```

dotaxis (approved):

davfsa (approved):
+1 from me. Thanks for making the PR!

I haven't really reviewed the TS part, ill leave that up to @Jas-SinghFSU  :)

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,95 @@ import Menu from 'src/components/shared/Menu';
...
+
+// Execute shell commands safely with path expansion
+const executeCommand = async (command: string) => {
+    try {
+        console.log('Executing command:', command);
+        const output = await execAsync(`/bin/bash -c '${command}'`);
+        console.log('Command output:', output);
+    } catch (err) {
+        console.error('Command failed:', command);
+        console.error('Error:', err);
```
davfsa, line 22: Also, we probably want to remove the console logs as to not spam users

davfsa (approved):

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,93 @@ import Menu from 'src/components/shared/Menu';
...
-                            const buttonClicked = event.get_button()[1];
+                return monitors.map((monitor) => {
+                    const sanitizedPath = getRecordingPath().replace(/"/g, '\\"');
 
-                            if (buttonClicked !== Gdk.BUTTON_PRIMARY) {
-                                return;
-                            }
+                    return (
+                        <MenuItem
+                            key={monitor.name}
```
Jas-SinghFSU, line 40:
```suggestion
```
Don't think `key` is a valid property AFAIK. You may be thinking of React component `key` property for mapping?

siddharthjain25: Yeah, I was confused.

Jas-SinghFSU:
:heart: :tada:

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
+const executeCommand = async (command: string): Promise<void> => {
+    try {
+        await execAsync(`/bin/bash -c '${command}'`);
+    } catch (err) {
+        console.error('Command failed:', command);
+        console.error('Error:', err);
+    }
+};
+
+// Monitor dropdown for screen selection
```
Jas-SinghFSU, line 24: Let's move these helper functions to `src/components/menus/dashboard/shortcuts/helpers.ts`

Jas-SinghFSU: If we're going to comment the functions, let's use the [JSDoc spec](https://jsdoc.app/about-getting-started).

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
-                            App.get_window('dashboardmenu')?.set_visible(false);
+                                App.get_window('dashboardmenu')?.set_visible(false);
 
-                            execAsync(`${SRC_DIR}/scripts/screen_record.sh start screen ${monitor.name}`).catch((err) =>
-                                console.error(err),
-                            );
-                        }}
-                    />
-                ));
+                                // Update command to match new script argument order
```
Jas-SinghFSU, line 46:
```suggestion
```
Noisy comment. Refer to section 4 of https://mitcommlab.mit.edu/broad/commkit/coding-and-comment-style/

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
+                    const sanitizedPath = getRecordingPath().replace(/"/g, '\\"');
+                    const command = `${SRC_DIR}/scripts/screen_record.sh start region "${sanitizedPath}"`;
+                    executeCommand(command);
                 }}
             />
         </Menu>
     );
 };
 
+// Recording button to start/stop screen recording
```
Jas-SinghFSU, line 70: Noisy comment. Adds nothing more than what the function name already explains.

Jas-SinghFSU, line 70:
```suggestion
```
Noisy comment. Adds nothing more than what the function name already explains.

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
-                const buttonClicked = event.get_button()[1];
+                if (event.get_button()[1] !== Gdk.BUTTON_PRIMARY) return;
 
-                if (buttonClicked !== Gdk.BUTTON_PRIMARY) {
-                    return;
-                }
+                const sanitizedPath = getRecordingPath().replace(/"/g, '\\"');
 
-                if (isRecording.get() === true) {
+                if (isRecording.get()) {
```
Jas-SinghFSU, line 82: Why are we making so many changes here just to refactor coding style? These aren't functional changes.

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
+                    const command = `${SRC_DIR}/scripts/screen_record.sh stop "${sanitizedPath}"`;
+                    executeCommand(command);
                 } else {
                     const monitorDropdownList = MonitorListDropdown() as Gtk.Menu;
                     monitorDropdownList.popup_at_pointer(event);
                 }
             }}
         >
-            <label className={'button-label txt-icon'} label={'󰑊'} />
+            <label className="button-label txt-icon" label="󰑊" />
```
Jas-SinghFSU, line 92: Why are we polluting the PR with just formatting changes?

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
+        console.error('Command failed:', command);
+        console.error('Error:', err);
+    }
+};
+
+// Monitor dropdown for screen selection
 const MonitorListDropdown = (): JSX.Element => {
     const monitorList: Variable<AstalHyprland.Monitor[]> = Variable([]);
 
-    const monitorBinding = Variable.derive([bind(hyprlandService, 'monitors')], () =>
```
Jas-SinghFSU, line 13: Lots of reformatting pollution here. Let's only make functional changes unless we're violating linter rules.

src/components/menus/dashboard/shortcuts/buttons/RecordingButton.tsx
```diff
@@ -4,79 +4,92 @@ import Menu from 'src/components/shared/Menu';
...
-        monitorList.set(hyprlandService.get_monitors()),
-    );
+    const monitorBinding = Variable.derive([bind(hyprlandService, 'monitors')], () => {
+        monitorList.set(hyprlandService.get_monitors());
+    });
 
     return (
-        <Menu className={'dropdown recording'} halign={Gtk.Align.FILL} onDestroy={() => monitorBinding.drop()} hexpand>
+        <Menu className="dropdown recording" halign={Gtk.Align.FILL} onDestroy={() => monitorBinding.drop()} hexpand>
```
Jas-SinghFSU, line 33: Reformatting pollution

Jas-SinghFSU:
Seems like we have a few linter errors.

Jas-SinghFSU:
Noice, let's merge this! Good work

PR #830, opened 2025-03-13
readme: add bc to required dependencies

dotaxis:
Add `bc` to required dependencies

https://github.com/Jas-SinghFSU/HyprPanel/commit/b1f38f62cca4ed0e99facff35c54cf1d537c1347 introduced the use of this but it is not included in Arch or Fedora by default.

AUR package will need updating as well.


Jas-SinghFSU:
Does the nix package need to be updated as well?

dotaxis:
> Does the nix package need to be updated as well?

Probably. I have never used nix and am not familiar with how it works. Does it just need to be added to the `flake.nix`?

Jas-SinghFSU:
I wonder if we even need `bc` here. Couldn't we just use `awk`? Since you have fractional scaling, can you do me a favor and replace
```sh
        scaled_width=$(echo "${w} / ${scale}" | bc)
        scaled_height=$(echo "${h} / ${scale}" | bc)
```
with
```sh
        scaled_width=$(awk "BEGIN {print $w / $scale}")
        scaled_height=$(awk "BEGIN {print $h / $scale}")
```
If it still works, then we can remove having to install a whole new dependency for floating point calculations.

dotaxis:
I don't actually have fractional scaling but hyprpanel reports 1 as 1.00. I can still test but I'm not sure if the results will mean much.

dotaxis:
`awk` does work on my machine so bc is not required.

Jas-SinghFSU:
Big W, thanks for testing!

PR #836, opened 2025-03-14
Nixos: Added bar.workspaces.applicationIconMap and bar.workspaces.workspaceIconMap

### LINKED ISSUE #840: workspace icons in home manager
could you make a way to give workspaces specific icons trough home manager, i think there is a way without homemanager, but at least i couldn't find a way to do it though home manager, but im configuring it through home manager as that is the way i like to configure it

Ifaen:
I noticed that these two options were missing in the settings of the nixos home-manager module. I tried to added them, but also I got into the issue that this function(?) was flattening the values:
```
flattenAttrs = attrSet: prefix:
    let
      process = key: value:
        if builtins.isAttrs value then
          flattenAttrs value "${prefix}${key}."
        else
          { "${prefix}${key}" = value; };
    in
      builtins.foldl' (acc: key:
        acc // process key attrSet.${key}
    ) {} (builtins.attrNames attrSet);
```

So, instead of getting: 
```
"bar.workspaces.applicationIconMap": {
  "class:cursor-url-handler": "󰨞", "class:vesktop": "󰙯"
},
```
I was getting:
```
"bar.workspaces.applicationIconMap.class:cursor-url-handler": "󰨞",
"bar.workspaces.applicationIconMap.class:vesktop": "󰙯",
```

So I changed `if builtins.isAttrs value then` to `if builtins.isAttrs value && !(key == "applicationIconMap" || key == "workspaceIconMap") then`.

To be sincere, I don't like it that much changing and adding those key exception, because is more like a workaround and not really a solution from me by having the keys being hardcoded. 

Then I noticed that the hyprpanel.layout are outside of hyprpanel.settings, which don't get proccesed by the flattenAttrs. Maybe is a better idea to have the mapping of Icons to its own separated submodule?

I apologize for not being brief with this, it's my first pull request and really want to contribute to this

Closes #840

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -344,13 +344,15 @@ in
       bar.windowtitle.truncation = mkBoolOption true;
       bar.windowtitle.truncation_size = mkIntOption 50;
       bar.workspaces.applicationIconMap = mkOption {
-        type = types.attrsOf types.str;
-        default = {};
-        example = {
-          "[dD]iscord" = "󰙯";
-          "title:YouTube" = "";
-          "class:wezterm$" = "";
-        };
+        type = jsonFormat.type;
+        default = null;
+        example = ''
+          {
+            "[dD]iscord" = "󰙯";
+            "title:YouTube" = "";
+            "class:wezterm$" = "";
+          }
+        '';
       };
       bar.workspaces.applicationIconEmptyWorkspace = mkStrOption "";
       bar.workspaces.applicationIconFallback = mkStrOption "󰣆";
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -5,21 +5,21 @@ let
   cfg = config.programs.hyprpanel;
 
   jsonFormat = pkgs.formats.json { };
-  
+
   # No package option
   package = if pkgs ? hyprpanel then pkgs.hyprpanel
   else abort ''
 
-  ********************************************************************************
-  *                                  HyprPanel                                   *
-  *------------------------------------------------------------------------------*
-  *                         You didn't add the overlay!                          *
-  *                                                                              *
-  * Either set 'overlay.enable = true' or manually add it to 'nixpkgs.overlays'. *
-  * If you use the 'nixosModule' for Home Manager and have 'useGlobalPkgs' set,  *
-  *                  you will need to add the overlay yourself.                  *
-  ********************************************************************************
-  '';
+        ********************************************************************************
+        *                                  HyprPanel                                   *
+        *------------------------------------------------------------------------------*
+        *                         You didn't add the overlay!                          *
+        *                                                                              *
+        * Either set 'overlay.enable = true' or manually add it to 'nixpkgs.overlays'. *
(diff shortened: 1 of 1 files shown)


nix/module.nix
```diff
@@ -5,47 +5,48 @@ let
...
 
-  # TODO: Please merge https://github.com/Jas-SinghFSU/HyprPanel/pull/497
-  #       Do not ask what these do...
   flattenAttrs = attrSet: prefix:
     let
       process = key: value:
-        if builtins.isAttrs value then
+        # If the value is an attribute set and the key is not one of the exceptions,
+        # recursively flatten the attribute set
+        if builtins.isAttrs value && !(key == "applicationIconMap" || key == "workspaceIconMap") then
```
Jas-SinghFSU, line 37: Does this mean that the condition here will have to be updated for each new json setting that we add? Can we make this more extensible?

Ifaen: This approach isn't as extensible as it could be because each setting that shouldn't be flattened would need to be explicitly hardcoded. That would go against the intent of [this PR](https://github.com/Jas-SinghFSU/HyprPanel/pull/497), which was designed to flatten everything inside cfg.settings.

Though with Layout there is another situation, because is not inside of cfg.settings and, near the end of the file, the code explicitly merges cfg.layout after flattening the other attributes. This prevents cfg.layout from being affected by flattenAttrs. The relevant code snippet is:
```
flatSet = flattenAttrs (lib.attrsets.recursiveUpdate cfg.settings theme) "";
mergeSet = flatSet // (flattenAttrs cfg.override "");
fullSet = if cfg.layout == null then mergeSet else mergeSet // cfg.layout;
finalConfig = toNestedObject fullSet;
```
Instead of keeping bar.workspaces.applicationIconMap and bar.workspaces.workspaceIconMap inside cfg.settings, we could move them to a new cfg.mapping attribute. This would allow for easier expansion in the future while avoiding conflicts with flattenAttrs.

Not trying to say that flattenAttrs is the problem, because is doing its job for the other settings properly. Is more like, in this specific case, there's a conflict in how the values should be structured

Would it be better to update this PR with the cfg.mapping change, or should I close this one and open a new PR from master? since I would be starting again and using another approach. Sorry for not doing it from the start, I just thought about this now that I took another look to the entire code hehe

davfsa:
I believe #854 has already handled most of what this pr set out to do.

If there is anything else left to add, then it will need a rebase and then orangci can have a look :)

Ifaen:
> I believe #854 has already handled most of what this pr set out to do.
> 
> If there is anything else left to add, then it will need a rebase and then orangci can have a look :)

Yeah, I am pretty sure is the same issue, don't know for sure because the user I assume uses a formatter like Alejandra which formatted all the code, so I can't be sure what's different from I did exactly, but is also good either way, because is easier to read and doesn't has that weird workaround I tried to do

It works for workspace icon mapping, but haven't tested for application icon mapping yet, if not I will do it from #854's code and will make another pr

Issue #829, opened 2025-03-12
NixOS - How to add the overlay when using nixosModule and having useGlobalPkgs to true


Lurgrid:
Hello, 

I'm new to the NixOS environment and I'd like to install Hyprpanel because I really like the project but I don't know how to add the overlay when using ``nixosModule``.

Error :
```
       ********************************************************************************
       *                                  HyprPanel                                   *
       *------------------------------------------------------------------------------*
       *                         You didn't add the overlay!                          *
       *                                                                              *
       * Either set 'overlay.enable = true' or manually add it to 'nixpkgs.overlays'. *
       * If you use the 'nixosModule' for Home Manager and have 'useGlobalPkgs' set,  *
       *                  you will need to add the overlay yourself.                  *
       ********************************************************************************
```

My flake :
```
{
  description = "NixOS Configuration";

  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
    hyprland = {
      url = "github:hyprwm/Hyprland";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    hyprpanel = {
      url = "github:jas-singhfsu/hyprpanel";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    home-manager = {
      url = "github:nix-community/home-manager";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    firefox = {
      url = "github:nix-community/flake-firefox-nightly";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = { nixpkgs, ... } @ inputs: 
  {
    nixosConfigurations.tantale = nixpkgs.lib.nixosSystem {
      specialArgs = { 
        inherit inputs; 
      };
      modules = [
        ./configuration.nix
        inputs.home-manager.nixosModules.home-manager {
          home-manager.useGlobalPkgs = true;
          home-manager.useUserPackages = true;
          home-manager.users.culottes = import ./home.nix;
          home-manager.backupFileExtension = "hm-backup";

          home-manager.extraSpecialArgs = {
            inherit inputs;
          };
        }
      ];
    };
  };
}
```

Do you have an idea?

Ifaen:
> you will need to add the overlay yourself.  


As the message says, when you use home-manager as a module and have useGlobalPkgs=true, you must add the overlay into nixpkgs.overlays to be able to use hyprpanel. 

[...]

PR #841, opened 2025-03-17
fix: use mkIf instead of optionals to provent empty overlays and avoid conflicts

NklsCh:
- Used `lib.mkIf` instead of `lib.optionals` to prevent empty overlay definitions
- Ensures compatibility with `home-manager.useGlobalPkgs = true`, avoiding overlay conflicts

Fixes issue #829


NklsCh:
Upon running `nix flake check` I found out that this line:

https://github.com/Jas-SinghFSU/HyprPanel/blob/master/flake.nix#L79

Is soon do be deprecated and should be replaced with `overlays.default`

Issue #848, opened 2025-03-18
[Bug]: Matugen overrides settings with default value but wont update


NklsCh:
**Describe the bug**
I had an issue where hyprpanel got bigger than i set it to be after an update (look at the screens)
This "bug" could be fixed after disabling matugen 

**To Reproduce**
Run newest version
Enable Matugen under Theming

**Expected behavior**
Do not change size

**Screenshots**
Matugen Enabled:
![Image](https://github.com/user-attachments/assets/19714f08-b91c-4b10-92e8-37607f3c36b0)

Matugen Disabled:
![Image](https://github.com/user-attachments/assets/4d58e5e9-a9f4-4ab5-920e-a08f0c87ec5c)

**Desktop (please complete the following information):**
 - Distribution: NixOS Unstable
 - Window Manager/Desktop Environment: Hyprland

**Additional context**

NklsCh:
I was able to pin point the root cause to the changes made in this commit https://github.com/Jas-SinghFSU/HyprPanel/commit/a949b346324b6405404595d8dfa3fcb4019dbcfd the changes in `src/scss/style.ts`

After this commit, boolean options (such as `theme.bar.transparent`) and numerical values (like `theme.font.size`) are not correctly processed in variables.scss. 

The new logic attempts to process all `theme.*` options. Boolean values (like `theme.bar.transparent`) are written directly into SCSS variables, which is not valid SCSS syntax. And `theme.font.size` (which defaults to "1.2rem") was incorrectly processed as a hex color, causing it to be overwritten or removed from `variables.scss`

### Potential Fixes:

```ts
if (isHexColor(initialValue) && matugenColors !== undefined) {
    const matugenColor = getMatugenHex(initialValue as HexColor, matugenColors);
    result.push(`$${name.replace('theme.', '').split('.').join('-')}: ${matugenColor};`);
} else {
    result.push(`$${name.replace('theme.', '').split('.').join('-')}: ${initialValue};`);
}
```

this fixes the issue with the strings. As for the boolean issue with still remains, i havent found a solution yet

kito30:
has this issues been resolved ?

NklsCh:
When the pr is merged, yes

kito30:
weird the problem still there

NklsCh:
Could you send screenshots?

kito30:
With matugen:
[...]

PR #849, opened 2025-03-18
Fix: Prevent Matugen to override user settings with config defaults

NklsCh:
Fixes issue #848 

Previously, the function used opt.initial to determine the theme variable values. However, opt.initial always holds the default value, causing user-defined settings to be overwritten when enabling Matugenized colors.

These changes to the function prioritize opt.value (if available) over opt.initial. This ensures that existing user configurations are preserved while still allowing Matugen colors to be applied correctly.

Co-Authored-by: @PhoenixGamer339

### DIFF
--- a/src/scss/style.ts
+++ b/src/scss/style.ts
@@ -55,14 +55,14 @@ async function extractMatugenizedVariables(matugenColors: MatugenColors): Promis
                 continue;
             }
 
-            const initialValue = opt.value ?? opt.initial;
+            const value = opt.value;
 
-            if (!isHexColor(initialValue) && matugenColors !== undefined) {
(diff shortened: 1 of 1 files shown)


davfsa (changes requested):
I believe I might have introduced this bug, so I am really sorry!

Pleas find bellow an improvement to your patch.

Again sorry!

Rexcrazy804:
> I believe I might have introduced this bug, so I am really sorry!
I had encountered this when I was building your pr like a couple days ago thought it was cause I the config.json spec changed and I was using a readonly symlink as well. Really appreciate your work tho live reloading config.json is great (just needa wait for this bug to be resolved xD)

Rexcrazy804:
Just built this pr and its back to working as usual :)

PR #854, opened 2025-03-23
feat: adding iconWorkspaceMap option

### LINKED ISSUE #840: workspace icons in home manager

Ciflire:
I discovered that there's a nix function `toJSON` that does what you were trying to do with custom functions, i tried not to modify too much of the existing code, there's probably some dead code at that point. 
However, using this function should make it so you can add every last option. For now I have tested it with workspaceIconsMap and it works well. I don't know every last option of hyprpanel, so i believe this PR should be a draft until I gathered (or anyone has listed) missing options that needed this

I also regrouped `theme` option under `config.settings.theme.name` and `layout` under  `config.settings.layout` so it's kind of a breaking change for HM module, i did not want to bother with merging attrSet (reason why there might be some dead code)

Closes #840

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -1,79 +1,72 @@
-self: { lib, pkgs, config, ... }:
+self:
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
 let
-  inherit (lib) types mkIf mkOption mkEnableOption;
+  inherit (lib)
+    types
+    mkIf
+    mkOption
+    mkEnableOption
+    ;
 
   cfg = config.programs.hyprpanel;
 
   jsonFormat = pkgs.formats.json { };
-  
-  # No package option
-  package = if pkgs ? hyprpanel then pkgs.hyprpanel
-  else abort ''
 
-  ********************************************************************************
-  *                                  HyprPanel                                   *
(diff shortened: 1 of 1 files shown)


davfsa:
Wondering if this is the same as #836 🤔

Ciflire:
The issue highlights the same issue as I did, the fact that the layout is out of the settings, this is kind of unconventional to me (but I could still be wrong). However it tries to correct the existing code whereas I used a built-in function and removed a lot of "homemade" code which maintainers will not have to try to understand anymore or maintain. This **should** allow to just add options without caring if it will be correctly translated to json.

davfsa:
Thanks for the explanation!

Ill leave this up to orangcl and jas then :)

Jas-SinghFSU:
@orangci Could I have you do a once-over for this one. Unfortunately I'm not much of an authority when it comes to Nix stuff.

@Ciflire Thank you for creating this btw.

ndowens:
Did font-size get removed? As I can no longer find it in the settings nor does it build in Nix due to option is missing?

PR #862, opened 2025-03-25
Fix: missing semicolon that breaks nix build

luke-wriglesworth:
There is a missing semicolon in the latest commit that prevents HyprPanel configurations from being built with nix.


Jas-SinghFSU:
Really ought to add a nix linter. Thanks

PR #866, opened 2025-03-27
Add optional extended tooltip for updates module

Mara-Dawn:
I was missing this functionality after swapping from waybar to HyprPanel, so i tried my hand at implementing it myself.

This will optionally show a list of updated packages when hovering the updates module. I set the toggle to off by default to not change existing configs. Here is a preview:

image

When disabled, it will show the old tool-tip same as before:

image

I tried to not break compatibility with any existing previous configs which might contain custom scripts. 

One thing i am not quite happy with is that i needed to add a sleep between pacmans _checkupdate_ calls, as it will not return anything if you use it again too quickly. I debated over changing more of the existing code to enable me calling it only once, but in the end I settled for the sleep as it results in the least amount of changes overall and works perfectly fine. Changing the behavior of the _checkUpdates.sh_ further would also likely break compatibility with existing user scripts.

I'm happy to implement a better way of doing it though if someone can come up with one.

Also this is my first open-source PR and my first time using TypeScript, so if i missed anything obvious please tell me as well!

### DIFF
--- a/scripts/checkUpdates.sh
+++ b/scripts/checkUpdates.sh
@@ -1,49 +1,87 @@
 #!/bin/bash
 
-check_arch_updates() {
-    official_updates=0
-    aur_updates=0
-	if command -v paru &> /dev/null; then
-		aur_helper="paru"
-	else
-		aur_helper="yay"
-	fi
+has_param() {
+    local term="$1"
+    shift
+    for arg; do
+        if [[ $arg == "$term" ]]; then
+            return 0
+        fi
+    done
+    return 1
+}
 
-    if [ "$1" = "-y" ]; then
-        aur_updates=$($aur_helper -Qum 2>/dev/null | wc -l)
-    elif [ "$1" = "-p" ]; then
-        official_updates=$(checkupdates 2>/dev/null | wc -l)
+check_arch_updates() {
+    if command -v paru &> /dev/null; then
+        aur_helper="paru"
     else
-        official_updates=$(checkupdates 2>/dev/null | wc -l)
-        aur_updates=$($aur_helper -Qum 2>/dev/null | wc -l)
+        aur_helper="yay"
     fi
 
-    total_updates=$((official_updates + aur_updates))
+    if has_param "-tooltip" "$@"; then
+        command=" | head -n 50"
+        official_updates=""
+        aur_updates=""
+        sleep 1 # if you use the checkupdates command again too quickly, it will not return anything
+    else
+        command="2>/dev/null | wc -l"
+        official_updates=0
+        aur_updates=0
+    fi
 
-    echo $total_updates
+    aur_command="$aur_helper -Qum $command"
+    official_command="checkupdates $command"
+
+    if has_param "-y" "$@"; then
+        aur_updates=$(eval "$aur_command")
+    elif has_param "-p" "$@"; then
+        official_updates=$(eval "$official_command")
+    else
+        aur_updates=$(eval "$aur_command")
+        official_updates=$(eval "$official_command")
+    fi
+
+    if has_param "-tooltip" "$@"; then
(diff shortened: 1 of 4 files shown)


scripts/checkUpdates.sh
```diff
@@ -1,49 +1,87 @@
...
+        aur_helper="paru"
+    else
+        aur_helper="yay"
+    fi
+
+    if has_param "-tooltip" "$@"; then
+        command=" | head -n 50"
+        official_updates=""
+        aur_updates=""
+        sleep 1 # if you use the checkupdates command again too quickly, it will not return anything
```
Jas-SinghFSU, line 25: Is this specifically in case someone is polling this every < 1 second?

Jas-SinghFSU:
Looks good otherwise, just need to fix a few linter issues in `src/components/bar/modules/updates/index.tsx`.

src/components/bar/modules/updates/index.tsx
```diff
@@ -29,15 +32,31 @@ const processUpdateCount = (updateCount: string): string => {
     return `${updateCount.padStart(2, '0')}`;
 };
 
+const processUpdateTooltip = (updateTooltip: string, updateCount: Variable<string>): string => {
+    let defaultTooltip = updateCount.get() + " updates available";
+    if (!extendedTooltip.get()) return defaultTooltip;
+    return defaultTooltip + "\n\n" + updateTooltip;
+};
```
Jas-SinghFSU, line 39:
```suggestion
    const defaultTooltip = updateCount.get() + ' updates available';
    if (!extendedTooltip.get()) return defaultTooltip;
    return defaultTooltip + '\n\n' + updateTooltip;
};
```

src/components/bar/modules/updates/index.tsx
```diff
@@ -29,15 +32,31 @@ const processUpdateCount = (updateCount: string): string => {
...
     processUpdateCount,
 );
 
+const tooltipPoller = new BashPoller<string, [Variable<string>]>(
+    pendingUpdatesTooltip,
+    [bind(extendedTooltip), bind(postInputUpdater), bind(updateTooltipCommand)],
+    bind(pollingInterval),
+    updateTooltipCommand.get(),
+    processUpdateTooltip,
+    pendingUpdates
```
Jas-SinghFSU, line 55:
```suggestion
    pendingUpdates,
```

scripts/checkUpdates.sh
Mara-Dawn: The checkupdate call gets used by both pollers back to back here, once to get the total number of updates and once to get the actual list.

scripts/checkUpdates.sh
```diff
@@ -1,49 +1,87 @@
...
+    local term="$1"
+    shift
+    for arg; do
+        if [[ $arg == "$term" ]]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
```
Jas-SinghFSU, line 13:
```suggestion

wait_for_process_to_finish() {
    local process_name="$1"
    while pgrep -a "$process_name" >/dev/null; do 
        sleep 0.1
    done
}

```
Let's implement a function that lets us do a more robust check than a static sleep value.

scripts/checkUpdates.sh
Jas-SinghFSU, line 25:
```suggestion
        wait_for_process_to_finish "checkupdates"
```
And we can just call the above function here so we're only waiting until needed.

Mara-Dawn:
Good idea! This is much cleaner

Jas-SinghFSU:
Sweet, thanks! :heart: :tada:

Mara-Dawn:
It seems the loop is not enough time between calls, i still often see the total number without pacman updates

image

We might have to go back to a fixed wait time

PR #870, opened 2025-03-29
Add option to ignore items in systray

ndowens:


ndowens:
I tested this and it does work for me to remove the nm-applet since I am using systemd-networkd

PR #871, opened 2025-03-30
Changed weather module location.

siddharthjain25:
Removed weather module from clock menu and added it to weather menu.


Jas-SinghFSU:
We should add a new menu for specifically the calendar (that we can extend) but shouldn't remove the calendar from the clock menu.

Jas-SinghFSU:
Hey @siddharthjain25, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #872, opened 2025-03-31
Fix: Stable update info polling

Mara-Dawn:
This is further optimization of #866. 

I rewrote the polling part so that it only polls once with both the normal and the extended tooltip setting. 

The previous implementation relied on waiting between two _checkupdate_ calls for a certain amount of time, as calling them back to back too quickly made the second call return nothing. This approach of waiting until the first process finished was often not enough to reliably get the second call to return data. Now there will only ever be one call per polling interval, ensuring that it will always work properly.

I also noticed the _stop()_ function of the _Poller_ class didn't actually stop the execution of the poller, so i added an _isActive_ flag to ensure the poller stops running.

### DIFF
--- a/scripts/checkUpdates.sh
+++ b/scripts/checkUpdates.sh
@@ -13,9 +13,13 @@ has_param() {
 
 wait_for_process_to_finish() {
     local process_name="$1"
-    while pgrep -a "$process_name" >/dev/null; do 
-        sleep 0.1
+    local pid
+
+    while pid=$(pgrep -x "$process_name"); do
+        wait "$pid" 2>/dev/null || true
     done
+
+    sleep 2
 }
 
 check_arch_updates() {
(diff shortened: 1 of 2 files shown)


Mara-Dawn:
alternatively i could also just go back to setting the delay to a fixed larger value, that seems to work as well

Jas-SinghFSU:
While I do want to move this to code instead of a script, I think having it purpose built in this way for Arch gets a bit messy. We probably want a dedicated service for each package manager/distro.

In the meantime, we can go back to a hybrid of the old approach which should be pretty safe:
```
wait_for_process_to_finish() {
    local process_name="$1"
    local pid

    while pid=$(pgrep -x "$process_name"); do
        wait "$pid" 2>/dev/null || true
    done

    sleep 2
}
```
This will wait for all processes of `checkupdates` to finish and then wait 2 seconds. This way you have a nice buffer on slow computers where `checkupdates` takes longer than 1 second, or 2 seconds. This checks for both the process AND provides a safety buffer.

Mara-Dawn:
You are right of course, I was not considering other package managers here. 

I Also didn't mean to close the PR, it just did that when i removed my previous commits.
I updated the `checkupdates.sh` and also fixed a small formatting issue in the tooltip.

Apologies, I might try my hand at implementing distro specific services in the future.

PR #897, opened 2025-04-12
feat: Add CPU and RAM usage graphs

make-42:


src/components/bar/modules/ram/index.tsx
```diff
@@ -25,19 +28,46 @@ const ramPoller = new FunctionPoller<GenericResourceData, [Variable<boolean>]>(
...
+    if (usage < 25) return '▂';
+    if (usage < 37.5) return '▃';
+    if (usage < 50) return '▄';
+    if (usage < 62.5) return '▅';
+    if (usage < 75) return '▆';
+    if (usage < 87.5) return '▇';
+    return '█';
+};
+
+const graphBinding = showGraph
```
Jas-SinghFSU, line 42: We can't use conditionals like this because once `showGraph`  changes, it won't be tracked unless put in a binding.

make-42: how do i fix this?

Jas-SinghFSU:
Hey @make-42, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #909, opened 2025-04-27
better runUpdates script for arch users

pabcihba:
better runUpdates script for arch users

### DIFF
--- a/scripts/runUpdates.sh
+++ b/scripts/runUpdates.sh
@@ -6,10 +6,19 @@ install_arch_updates() {
     echo "Updating Arch Linux system..."
     sudo pacman -Syu
     echo "Updating AUR packages..."
+    if command -v paru &> /dev/null; then
     paru -Syu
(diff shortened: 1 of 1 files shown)


scripts/runUpdates.sh
```diff
@@ -6,7 +6,13 @@ install_arch_updates() {
     echo "Updating Arch Linux system..."
     sudo pacman -Syu
     echo "Updating AUR packages..."
+    if command -v paru &> /dev/null; then
     paru -Syu
+    elif command -v yay &> /dev/null; then
+    yay -Syu
+    else
+    echo "Missing AUR Helper. Try installing yay or paru"
```
davfsa, line 14:
```suggestion
    if command -v paru &> /dev/null; then
        paru -Syu
    elif command -v yay &> /dev/null; then
        yay -Syu
    else
        echo "Missing AUR Helper. Try installing yay or paru"
```

PR #937, opened 2025-05-13
Updated hourly weather time to respect "military" clock setting

Tawank:
This PR updates the `HourlyTime` component to respect the `options.menus.clock.time.military` setting for time formatting.
It looks like this, if you have set `"menus.clock.time.military": true`:
image


Jas-SinghFSU:
:tada: Ty!

PR #951, opened 2025-05-23
refactor(bar): :art: generate widgets map dynamically from components

KernelDiego:
Replaces hardcoded widget map definition with a dynamic generation using Object.entries, allowing for consistent wrapping logic and special-case handling for Workspaces and ModuleSeparator.

### DIFF
--- a/src/components/bar/exports.ts
+++ b/src/components/bar/exports.ts
@@ -30,15 +30,15 @@ import { WorldClock } from '../../components/bar/modules/worldclock/index';
 import { ModuleSeparator } from './modules/separator';
 
 export {
-    Menu,
+    Menu as Dashboard,
     Workspaces,
-    ClientTitle,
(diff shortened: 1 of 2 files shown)


KernelDiego:
Also add an optimization that stops the recording poller when shortcuts are disabled, avoiding redundant polling and reducing resource usage.
and reducing resource usage.

Jas-SinghFSU:
No idea why there's a conflict since nothing in this file changed in `master` lol.

KernelDiego:
Yes, I just realized that both changes were made almost at the same time. Let me make another PR fixing it.

Jas-SinghFSU:
I just realized that in my new refactor of #934 that I don't export and instead added a factory function.

But I've cherrypicked your commit for fixing the redundant polling into that PR so you get the commit and code credit :smile: .

Closing this one.

PR #956, opened 2025-05-26
Use only JSON conversion

### LINKED ISSUE #886: New nix/home-manager module behaviour breaks theming use cases
Since #854 was merged, the UX of specifying more advanced theming through the HM module has become convoluted and inconsistent. This is presumably due to the fact that on one hand the module still hard-codes allowed option values, but on the other hand leverages JSON conversion.

This means that we cannot include e.g. `settings.theme.bar.border.color` because it's not allowed, so we are forced to do this in `override`. However, likely due to the merging behaviour of the attribute set resulting from the JSON conversion and that from the `override`, using `override` causes valid theme-settings from the `settings` attrset to be ineffective (probably overwritten by empty values, or the fields being deleted alltogether?)

Probably a better functionality of the HM module would be given if a choice were made to *either* do straight JSON conversion at the cost of removing option checking, *or* declare all in full detail. In the latter case, one could probably still use a "class/type" approach for each "submodule", so one could use one submodule type with standard fields for all supported properties of a certain bar element.

In general, for a good UX/DX one would which for a consistent and generic options for the theme on the level of each element type (e.g. `button`), which, if set, constitutes the default for all, unless overridden for a specific instance.

For reference here my verbatim config module for `hyprpanel`, which in this state doesn't yet completely work as intended, and makes for a pretty bad DX due to the fact that hierarchical defaults are ineffective.

```nix
{
  inputs,
  config,
  osConfig,
  lib,
  ...
}: let
  colorScheme = config.lib.stylix.colors;
  stylixColor = n: "#${colorScheme.${n}}";

  applyMultiPropsTo = basePath: props: targets: let
    # Helper function to create a single property for one target
    makeProperty = target: propName: propValue: let
      # Construct the full attribute path
      fullPath = "${basePath}.${target}.${propName}";
    in
      # Return a singleton attrset with the constructed path
      {"${fullPath}" = propValue;};

    # Generate properties for a single target
    makeTargetProps = target:
      builtins.foldl' (
        acc: prop:
        # Merge the accumulator with the new property
          acc // (makeProperty target (builtins.head prop) (builtins.elemAt prop 1))
      )
      {}
      (builtins.attrValues (builtins.mapAttrs (name: value: [name value]) props));

    # Combine properties for all targets
    allProps =
      builtins.foldl' (
        acc: target:
        # Merge accumulated properties with new target properties
          acc // (makeTargetProps target)
      )
      {}
      targets;
  in
    allProps;
in {
  imports = [inputs.hyprpanel.homeManagerModules.hyprpanel];

  programs.hyprpanel = {
    # Enable the module.
    # Default: false
    enable = true;
    overlay.enable = true;

    # Automatically restart HyprPanel with systemd.
    # Useful when updating your config so that you
    # don't need to manually restart it.
    # Default: false
    # NOTE: now obsolete
    # systemd.enable = true;

    # Add '/nix/store/.../hyprpanel' to your
    # Hyprland config 'exec-once'.
    # Default: false
    hyprland.enable = true;

    # Fix the overwrite issue with HyprPanel.
    # See below for more information.
    # Default: false
    overwrite.enable = true;

    # Override the final config with an arbitrary set.
    # Useful for overriding colors in your selected theme.
    # Default: {}
    # instead of direct stylix support (TODO?) we just assign here from the stylix palette
    override =
      applyMultiPropsTo "theme.bar.buttons"
      {
        border = stylixColor "brown"; # this is actually ~magenta in outrun, but not "magenta" (which is redder)
        enableBorder = true;
        text = stylixColor "bright-blue";
        icon = stylixColor "cyan";
        icon_background = stylixColor "base03";
      } [
        "battery"
        "bluetooth"
        "clock"
        "dashboard"
        "media"
        "modules.cpu"
        "modules.hypridle"
        "modules.submap"
        "network"
        "notifications"
        "systray"
        "volume"
        "windowtitle"
        "workspaces"
      ]
      // {
        theme = {
          font = {
            inherit (config.stylix.fonts.sansSerif) name;
            label = config.stylix.fonts.sansSerif.name;
            # size = "${toString (builtins.ceil (config.stylix.fonts.sizes.applications * 1.4))}px";
            size = "1.1rem";
            weight = 400;
          };
          bar = {
            background = stylixColor "orange";
            shadow = "0 0 0 0";
            shadowMargins = "0 0 0 0";
            buttons = {
              radius = "0.8em";
              borderSize = "0.07em";
              dashboard.background = stylixColor "base02";
              notifications.total = stylixColor "cyan";
              clock.text = stylixColor "cyan";
              workspaces = {
                occupied = stylixColor "cyan";
                available = stylixColor "base04";
                active = stylixColor "magenta";
                hover = stylixColor "magenta";
                numbered_active_highlighted_text_color = stylixColor "magenta";
                numbered_active_underline_color = stylixColor "cyan";
              };
            };
          };
        };
      };
    # theme.bar.buttons.icon_background = stylixColor "base03";
    # theme.bar.buttons.dashboard = {
    #   background = stylixColor "base02";
    #   icon = stylixColor "bright-blue";
    # };

    # Configure and theme almost all options from the GUI.
    # Options that require '{}' or '[]' are not yet implemented,
    # except for the layout above.
    # See 'https://hyprpanel.com/configuration/settings.html'.
    # Default: <same as gui>
    settings = {
      wallpaper.enable = false;
      scalingPriority = "hyprland";

      # Configure bar layouts for monitors.
      # See 'https://hyprpanel.com/configuration/panel.html'.
      # Default: null
      layout = {
        "bar.layouts" = {
          "*" = {
            left = [
              "dashboard"
              "hypridle"
              "workspaces"
              "submap"
              "media"
            ];
            middle = ["windowtitle"];
            right =
              ["cpu"]
              ++ lib.optionals osConfig.networking.networkmanager.enable ["network"]
              ++ [
                "bluetooth"
                "systray"
                "volume"
                "notifications"
              ]
              ++
              # primitive way to check whether we're on a laptop
              lib.optionals (osConfig.services.tlp.enable || osConfig.services.upower.enable) ["battery"]
              ++ [
                "clock"
              ];
          };
        };
      };

      theme = {
        # Import a theme from './themes/*.json'.
        # Default: ""
        name = "tokyo_night_split";

        bar = {
          floating = false;
          transparent = true;
          # opacity = 15;
          border = {
            # color = stylixColor "brown"; # this is actually ~magenta in outrun, but not "magenta" (which is redder)
            width = "0";
            location = "none";
          };
          outer_spacing = "0";
          margin_top = "0.1em";
          margin_bottom = "0.1em";
          margin_sides = "0.0em";
          dropdownGap = "2.7em";

          menus = {
            monochrome = true; # will make all colours the same per default
            border.radius = "0.8em";
            tooltip.radius = "0.8em";
            popover.radius = "0.8em";
            menu = {
              notifications = {
                height = "48em";
                # width = "12em"; # FIXME: would be good to have this option
              };
            };
            menu.clock.scaling = 80;
            menu.media.scaling = 90;
          };

          buttons = {
            # text = stylixColor "bright-blue";
            # icon = stylixColor "cyan";
            # icon_background = stylixColor "base03";
            enableBorders = true;
            monochrome = true;
            opacity = 10; # probably mutually exclusive with transparent?
            # transparent = true;
            borderSize = "0.07em";
            radius = "0.8em";
            spacing = "0.3em";
            innerRadiusMultiplier = "0.9";
            y_margins = "0.1em";
            workspaces = {
              fontSize = "1.0em";
              pill.height = "3em";
            };
            # dashboard = {
            #   background = stylixColor "base02";
            #   icon = stylixColor "bright-blue";
            # };
          };
        };

        notification = {
          border_radius = "0.8em";
        };

        osd = {
          opacity = 85;
          location = "center";
          orientation = "horizontal";
          radius = "0.8em";
        };
      };

      bar = {
        launcher.autoDetectIcon = true;
        workspaces = {
          show_numbered = true;
          show_icons = ! config.programs.hyprpanel.settings.bar.workspaces.show_numbered;
          reverse_scroll = true;
        };
        windowtitle = {
          class_name = false;
          custom_title = false;
          truncation_size = 35;
        };
        notifications.show_total = true;
        clock = {
          format = "%a %-d %b %H:%M W%V";
        };
        volume = {
          scrollDown = "hyprpanel vol +5";
          scrollUp = "hyprpanel vol -5";
        };
        customModules = {
          cpu.leftClick = "ghostty --class=ghostty.floatingLarge -e btop";
          hypridle = {
            label = false;
            offIcon = "⏳";
            onIcon = "🚀";
          };
          submap = {
            enabledIcon = "⌥";
            disabledText = "🔡"; # FIXME: has no effect
          };
        };
      };

      menus = {
        clock = {
          time = {
            military = true;
            hideSeconds = false;
          };
          weather.unit = "metric";
        };

        dashboard = {
          directories.enabled = false;
          shortcuts.enabled = false;
          stats.enable_gpu = false;
          powermenu = {
            avatar.image = "${toString config.xdg.dataFile."avatar.png".source}";
          };
        };
      };
    };
  };
}

```

DADA30000:
Should fix #886 
Now you can define every needed setting using NixOS options, and it will automatically override theme settings, and also defaults, no need for `override` option now.
But I can't test it right now, so idk if this actually works
For anyone who wants to test, just replace hyprpanel input in flake.nix like this:
Replace this
```
inputs = {
  hyprpanel.url = "github:Jas-SinghFSU/Hyprpanel";
}
```
with
```
inputs = {
  hyprpanel.url = "github:DADA30000/Hyprpanel";
}
```


DADA30000:
Also docs need to be changed accordingly

PR #963, opened 2025-05-28
Use only JSON conversion

DADA30000:
#956


Jas-SinghFSU:
Hey @DADA30000, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #964, opened 2025-05-28
Add devShell to flake.nix (also update flake.lock)

DADA30000:
flake.lock wasn't updated in 4 months, there could be some vulnerabilities because of that


DADA30000:
> Even though the astal bindings seem to exist I still get these errors for gi (whatever that is): image Is there a way to fix this aswell?

nope, npx (or npmx I don't remember) doesn't want to generate bindings for AstalHyprland for some reason
It is possible tho to just get those bindings from someone else and just put them there manually

Septias:
There is really no way to generate them myself? xD

PR #981, opened 2025-06-02
fix: initialize media players on startup

davfsa:
They were previously not being loaded until they were seen for the first time

### DIFF
--- a/src/services/media/index.ts
+++ b/src/services/media/index.ts
@@ -57,6 +57,10 @@ export class MediaPlayerService {
 
         this.mediaTitle.set(noMediaText.get());
 
+        for (const player of this._mprisService.players) {
+            this._handlePlayerAdded(player);
+        }
+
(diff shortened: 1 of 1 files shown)


src/services/media/index.ts
```diff
@@ -57,6 +57,10 @@ export class MediaPlayerService {
 
         this.mediaTitle.set(noMediaText.get());
 
+        for (const player of this._mprisService.players) {
```
Jas-SinghFSU, line 60:
```suggestion
        for (const player of this._mprisService.get_players()) {
```
For list type properties let's use getter methods due to some issue with Vala and GJS: https://aylur.github.io/astal/guide/typescript/faq#error-can-t-convert-non-null-pointer-to-js-value

davfsa: I'll keep that in mind for next time!

Thanks!

PR #984, opened 2025-06-02
chore: simplify monitor matching by just checking geometry

davfsa:
A lot of the code I believe could be simplified, but didnt bother with that for now


Jas-SinghFSU:
To me this looks like it would add a regression in matching monitors where GDK scaling is different than Hyprland scaling, hence throwing the pixel coordinates off, no?

Jas-SinghFSU:
Hey Dav, we're hanging out in [Wayle](https://github.com/wayle-rs/wayle) as you may already be aware. Closing all the existing PRs on HyprPanel. See you on the other side.

PR #985, opened 2025-06-02
feat: add ability to specify connector as layout name and extend other layouts

davfsa:
Some example configuration of what is now possible:
```json
  "bar.layouts": {
    "base": {
      "left": ["workspaces"],
      "middle": ["cava"],
      "right": [
        "volume",
        "network",
        "bluetooth",
        "clock",
        "notifications",
        "hypridle",
        "power"
      ]
    },
    "DP-2": {
      "extends": "base"
    },
    "eDP-1": {
      "extends": "base",
      "right": [
        "volume",
        "network",
        "bluetooth",
        "battery",
        "clock",
        "notifications",
        "hypridle",
        "power"
      ]
    },
    "*": {
      "left": ["workspaces"],
      "right": ["volume", "clock", "notifications"]
    }
  },
```

This configuration will reuse the same "base" layout for both `eDP-1` (connector of my laptop display) and `DP-2` (my main display for my desktop), and I can override the right part of the layout for `eDP-1` to add the battery module.

### DIFF
--- a/src/components/bar/utils/monitors/index.ts
+++ b/src/components/bar/utils/monitors/index.ts
@@ -3,6 +3,13 @@ import { BarLayout, BarLayouts } from 'src/lib/options/types';
 import { GdkMonitorService } from 'src/services/display/monitor';
 import { MonitorMapping } from './types';
 import { JSXElement } from 'src/core/types';
+import AstalHyprland from 'gi://AstalHyprland?version=0.1';
+
+const emptyBar = {
+    left: [],
+    middle: [],
+    right: [],
+};
 
 /**
  * Returns the bar layout configuration for a specific monitor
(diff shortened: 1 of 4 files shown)


Jas-SinghFSU:
Love this, sorry about the delay. Reviewing now...

src/lib/options/types.ts
```diff
@@ -46,6 +46,7 @@ export type BarLayout = {
     left: BarModule[];
     middle: BarModule[];
     right: BarModule[];
+    extends: string?;
```
Jas-SinghFSU, line 49:
```suggestion
    extends?: string;
```

src/components/bar/utils/monitors/index.ts
```diff
@@ -12,24 +19,80 @@ import { JSXElement } from 'src/core/types';
...
-        return layouts[matchingKey];
-    }
+    let left = rootLayout.left;
+    let middle = rootLayout.middle;
+    let right = rootLayout.right;
 
-    if (wildcard) {
-        return layouts[wildcard];
+    let layout = rootLayout;
+    const visited = new Array([rootKey]);
```
Jas-SinghFSU, line 29:
```suggestion
    const visited = [rootKey];
```
Constructor syntax not needed. The current implementation creates an array of arrays `[["DP-1"]]`. And a single number passed in the `new Array(someVal)` if it ends up being a number creates an array of that size, rather than an array with that element.

The beauty of this awful language lol.

src/components/bar/utils/monitors/index.ts
```diff
@@ -12,24 +19,80 @@ import { JSXElement } from 'src/core/types';
...
+    while (
+        layout.extends !== undefined &&
+        (left === undefined || middle === undefined || right.length === undefined)
```
Jas-SinghFSU, line 32:
```suggestion
        (left === undefined || middle === undefined || right === undefined)
```
I imagine the `.length` was accidental?

Jas-SinghFSU (changes requested):

src/components/bar/utils/monitors/index.ts
davfsa: It was a leftover of a test, nice catch, ty!

Jas-SinghFSU (approved):

PR #992, opened 2025-06-03
Fix: reliably set GIO_EXTRA_MODULES for devShell and executable

geri1701:
My previous PR ([8422c6b8](https://github.com/Jas-SinghFSU/HyprPanel/commit/8422c6b80526f8289a30b93cb5b354d9f007141d)) only set GIO_EXTRA_MODULES in the devShell, which was not enough—running the hyprpanel executable outside the devShell could still fail due to missing environment variables.

This PR uses Nix’s wrapProgram to ensure GIO_EXTRA_MODULES is always set, whether running in the devShell or directly.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,6 @@
 {
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
-    
     astal = {
       url = "github:aylur/astal";
       inputs.nixpkgs.follows = "nixpkgs";
@@ -13,12 +12,7 @@
   };
 
   outputs =
-    {
-      self,
-      nixpkgs,
-      astal,
-      ags,
-    }:
+    { self, nixpkgs, astal, ags, ... }:
     let
       systems = [
         "x86_64-linux"
@@ -76,7 +70,6 @@
           swww
           pywal
         ]);
-
     in
     {
       devShells = forEachSystem (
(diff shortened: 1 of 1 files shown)


DADA30000:
> > ah wait, no, that wrapper had a check, maybe it was placed in here for a reason?
> 
> The wrapper attribute is still needed so users can continue to follow the instructions in the README and NixOs setup. I’ve added it back to ensure everything works as described. Thanks for your feedback!

no, I mean look at the previous wrapper script
if [ "$#" -eq 0 ]; then
                exec ${self.packages.${pkgs.stdenv.system}.default}/bin/hyprpanel
            else
                exec ${ags.packages.${pkgs.stdenv.system}.io}/bin/astal -i hyprpanel "$@"
            fi
you see a check here? It checks if there are arguments passed, and if there are, it does different thing, are you using AI?

geri1701:
I removed the check because it seems HyprPanel now handles CLI commands internally, and my focus here was on ensuring `GIO_EXTRA_MODULES` is set. I haven’t had a chance to audit the entire codebase yet. Do you think this check is still necessary?

Jas-SinghFSU:
Hyprpanel does leverages Astal to run CLI commands. This script does two things:
1. It launches Hyprpanel - which creates a hyprpanel instance in Astal
2. It checks if arguments are passed; if so, it sends CLI commands via Astal to the hyprpanel instance (started by step 1)

We need these because Astal is the one listening to CLI commands and forwarding them to our running instance.

geri1701:
Thanks for the clarification! I will add that again.

Issue #1010, opened 2025-06-17
`hyprpanel vol -5` not working as expected


kkanden:
The default scroll-down behavior on the volume component doesn't work - i can only increase the volume. when running `hyprpanel vol -5` from the terminal i get `error: Unknown option -5`.

 - Distribution: NixOS
 - Window Manager/Desktop Environment: Hyprland

L-Acacia:
(gjs:28979): hyprpanel-CRITICAL **: 12:36:26.848: Error running command "hyprpanel vol -5": Gio.IOErrorEnum: \u001b[31merror: \u001b[0mUnknown option -5)

```nix
pkgs = import nixpkgs {
  inherit system;
  overlays = overlays;
};

...

home.packages = with pkgs; [
  hyprpanel
]
```

I have the same issue using the provided flake and installing the package with the overlay.

jtliang24:
From what I can tell, the issue is that -5 is treated as another option, while +5 is properly treated as the argument.

kkanden:
closing this as the mentioned commited fixed the issue

PR #1022, opened 2025-06-23
Fix for volume down issue #1010

jtliang24:
Fixed the scrolling on the volume bar (as per #1010 ) not working correctly because it was interpreting a negative value as an option instead of an argument.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -325,8 +325,8 @@ in
       bar.volume.label = mkBoolOption true;
       bar.volume.middleClick = mkStrOption "";
       bar.volume.rightClick = mkStrOption "";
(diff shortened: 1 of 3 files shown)


Jas-SinghFSU:
I think the fix was to change the parameter expansions in the wrapper package from 
```
                exec ${ags.packages.${pkgs.stdenv.system}.io}/bin/astal -i hyprpanel "$@"
```
to
```
                exec ${ags.packages.${pkgs.stdenv.system}.io}/bin/astal -i hyprpanel "$*"
```

Can you verify on latest, if volume change works as expected?

jtliang24:
Yes this seems to fix the issue. I am closing the pull request.

Issue #1023, opened 2025-06-24
NixOS / Home-Manager no longer builds


smashedbanan:
**Describe the bug**
After updating my nix flake, the hyprpanel input complains about:
```
The option `home-manager.users.user.programs.hyprpanel.settings' in module `/nix/store/xrc7yj8kmljp7wx8w0ff1bdadh8v5jj6-source/modules/programs/hyprpanel' would be a parent of the following options, but its type `JSON value' does not support nested options.`
```

No changes to my hyprpanel.nix were made between flake updates:
```
• Updated input 'hyprpanel':
    'github:Jas-SinghFSU/HyprPanel/d563cdb1f6499d981901336bd0f86303ab95c4a5?narHash=sha256-oREAoOQeAExqWMkw2r3BJfiaflh7QwHFkp8Qm0qDu6o%3D' (2025-06-20)
  → 'github:Jas-SinghFSU/HyprPanel/2c0c66a8ddcdff69b2ff694fefbcd8df4387a9e7?narHash=sha256-wn6XV23Sz/nsDPI%2B1FGjtyMfAWJWcXRpNcryS47DJdY%3D' (2025-06-22)
```

**To Reproduce**
update nix flake via `nix flake update` and rebuild system via `nh` or `nixos-rebuild`
both update methods result in the evaluation failure of any `hyprpanel.settings` attribute.

**Expected behavior**
correct build of hyprpanel configuration

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - Distribution: NixOS & Home-Manager via Flakes
 - Window Manager/Desktop Environment: Hyprland

**Additional context**

Additional error output:

```
        - option(s) with prefix `home-manager.users.user.programs.hyprpanel.settings.bar' in module `/nix/store/d51f23ikygajx5gq7lb5w26kdh93hr9d-source/home-manager/programs/hyprPanel.nix'
        - option(s) with prefix `home-manager.users.user.programs.hyprpanel.settings.dummy' in module `/nix/store/d51f23ikygajx5gq7lb5w26kdh93hr9d-source/home-manager/programs/hyprPanel.nix'
        - option(s) with prefix `home-manager.users.user.programs.hyprpanel.settings.hyprpanel' in module `/nix/store/d51f23ikygajx5gq7lb5w26kdh93hr9d-source/home-manager/programs/hyprPanel.nix'
[...]

PR #1024, opened 2025-06-24
nix: migrate to downstream infrastructure

PerchunPak:
I have ported hyprpanel completely to nixpkgs, which includes home-manager module. This migrates everything to use it, so you don't have to maintain the same thing (and looking at the readme, you already have some issues with that)

The new code is available [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/hy/hyprpanel/package.nix) for the package build and [here](https://github.com/nix-community/home-manager/blob/master/modules/programs/hyprpanel/default.nix) for the module.

Fixes https://github.com/Jas-SinghFSU/HyprPanel/issues/1023

## Requirements

**Do not merge this until everything here is done**

- [x] https://github.com/NixOS/nixpkgs/pull/379994 
    hits nixos-unstable ([tracker](https://nixpk.gs/pr-tracker.html?pr=379994))
- [x] https://github.com/NixOS/nixpkgs/pull/419606
    hits nixos-25.05 ([tracker](https://nixpk.gs/pr-tracker.html?pr=419606))

- [x] https://github.com/nix-community/home-manager/pull/7303
- [x] https://github.com/nix-community/home-manager/pull/7321

For now, you can try it out using `nix build . --override-input nixpkgs ~/dev/nixpkgs/`


flake.nix
```diff
@@ -1,158 +1,75 @@
...
-        in
-        {
-          hyprpanel = self.packages.${pkgs.stdenv.system}.wrapper;
+          GIO_EXTRA_MODULES = "${pkgs.glib-networking}/lib/gio/modules";
         };
+      }
+    )
+    // {
+      overlay = builtins.warn ''
+        Overlay has been removed, because hyprpanel is not packaged in
```
DADA30000, line 65: "not"? Maybe you meant "now"?

DADA30000:
also I don't want the flake to be completely removed, I still want to use flake sometimes, to have latest changes, like look at hyprland, they have the package in nixpkgs, but they still have their flake, this is because changes can take week or two to get to the nixos-unstable, and waiting this much to get a bugfix is kinda annoying...
(sure you could do override, but meh)

DADA30000:
hm part can be completely removed, but package should stay

flake.nix
PerchunPak: Right, fixed

DADA30000:
> > (sure you could do override, but meh)
> 
> This is exactly what I am doing in this PR, you can still fetch the latest available version by using flake in this repo. The package here is not deprecated, it is just moved to nixpkgs to reduce the maintenance load
> 
> https://github.com/Jas-SinghFSU/HyprPanel/blob/3f086879/nix/overlay.nix#L4

ooh, I see now, just needed to read the code with more attention, sorry for being annoying :P
Although I personally would like to avoid overlays, afaik they can make some impact on eval time, or RAM usage while evaluationg, and overrides do not do that, but other than that, lgtm

PerchunPak:
> sorry for being annoying :P

No problem

> Although I personally would like to avoid overlays, afaik they can make some impact on eval time, or RAM usage while evaluationg, and overrides do not do that

Overrides and overlays are the same thing. Also, the performance impact from overlays is negligible: the most of the evaluation time is taken by NixOS modules. And in this case, it might even improve performance, because evaluation inside nixpkgs is cached more efficiently, than custom flakes, for some reason (just something I observed, not sure why though)

PR #1027, opened 2025-06-27
feat: configurable exclusivity

LoricAndre:
This implements a setting for the bar's exclusivity, instead of automatically computing it from the visibility variable.

As far as I can tell, the `BarLayout._visibilityVar` is toggled by an empty bar, which should (?) basically never happen.

My use case is that I have a keybind to show the bar only when I press `Super`. Without this, my windows will move each time the bar is shown. With this setting, I can show the bar over my other windows.

Please tell me if this needs further explanation and/or documentation.


Jas-SinghFSU:
Hey @LoricAndre, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #1058, opened 2025-07-13
feat: add ability to auto dnd on screenshare

CakeWithDivinity:
This adds an option that automatically turns on do-not-disturb in the astal notifd notification daemon when a screenshare starts (see the `screencast` event on https://wiki.hypr.land/0.47.0/Plugins/Development/Event-list/) and disables it, when the screenshare stops


Jas-SinghFSU:
Hey @CakeWithDivinity, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #1060, opened 2025-07-13
Fix region recording on rotated monitors

dotaxis:
Requires wf-recorder to be patched with -t flag

Will see if https://github.com/ammen99/wf-recorder/pull/313 goes through


Jas-SinghFSU:
Hey dot, as you know we've moved over to Wayle, so I'm closing the pending PRs for HyprPanel.

PR #1067, opened 2025-07-21
fix: netstat module ignoring rateUnit and round settings

### LINKED ISSUE #1015: netstat module doesn't respect rateUnit
**Describe the bug**
Module netstat does not respect rateUnit set. Eg, I've set it to "MiB" but it is always in auto

**To Reproduce**
Set customModules.netstat.rateUnit to "GiB", "MiB", "KiB" and it will always be in "auto"  

**Expected behavior**
Unit should show in specified unit

**Screenshots**
Image

 - Distribution: NixOS
 - Apparently there is also one reported instance of same behavior happening on Arch on discord

basnijholt:
The netstat module was creating two separate NetworkUsageService instances:
1. One in setupNetworkServiceBindings() that received the configuration settings
2. Another in the main module that actually displayed the data

This caused the rateUnit and round settings to be applied to the wrong service instance, resulting in network speeds always showing as "bytes/s" with 2 decimal places regardless of user configuration.

Fixed by:
- Moving NetworkUsageService instantiation before setupNetworkServiceBindings()
- Passing the service instance as a parameter to setupNetworkServiceBindings()
- Removing the redundant service creation in the helper function

Now the same service instance receives both the settings and displays the data, ensuring rateUnit (GiB/MiB/KiB/auto) and round settings work correctly.

Fixes #1015


Jas-SinghFSU:
Good catch! Just needs a quick linter fix
```
/home/runner/work/HyprPanel/HyprPanel/src/components/bar/modules/netstat/helpers.ts
Error:   8:1  error  Delete `⏎`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

Error: Process completed with exit code 1.
```

PR #1068, opened 2025-07-21
fix: netstat module showing 10x slower speeds than actual with (default) fast polling

basnijholt:
## Problem

When using fast polling intervals (like 100ms, which is actually the default!), the netstat module reports network speeds that are ~10x slower than actual speeds measured by external tools like speedtest.net. This creates a jarring disconnect where:

- Speed test shows: 300 Mbps download
- HyprPanel netstat shows: ~30 Mbps (4 MiB/s instead of 36+ MiB/s)
- Users cannot trust the module's accuracy for monitoring real network performance

## Root Cause

The issue was in the time difference calculation at line 76:

```typescript
// BROKEN: Forces minimum 1-second intervals even for fast polling
const timeDiff = Math.max((currentTime - this._previousNetUsage.time) / 1000, 1);
```

This "safety" code was designed to prevent division by zero, but it artificially inflated time intervals for fast polling scenarios:

### Example with 100ms polling interval:
- **Actual time between measurements**: 0.1 seconds
- **Forced minimum by old code**: 1.0 seconds
- **Rate calculation**: `bytes_transferred / time_interval`
- **Result**: `bytes_transferred / 1.0s` instead of `bytes_transferred / 0.1s`
- **Effect**: **Speeds appear 10x slower than reality**

### Real-world impact:
During a 300 Mbps speed test with 100ms polling:
- **Bytes transferred in 100ms**: ~3.75 MB
- **Correct calculation**: 3.75 MB ÷ 0.1s = **37.5 MB/s** (≈300 Mbps) ✅
- **Broken calculation**: 3.75 MB ÷ 1.0s = **3.75 MB/s** (≈30 Mbps) ❌

## Solution

Changed the minimum time interval from 1 second to 1 millisecond:

```typescript
// FIXED: Allows accurate sub-second polling while preventing division by zero
const timeDiff = Math.max((currentTime - this._previousNetUsage.time) / 1000, 0.001);
```

## Testing

Manual verification confirmed the fix:
- Before: 300 Mbps speed test → 4 MiB/s in module (10x error)
- After: 300 Mbps speed test → 36+ MiB/s in module (accurate)


basnijholt:
@Jas-SinghFSU, seems CI passes and this fixes a bug that will be present for everyone using the netstat module with defaults.

PR #1069, opened 2025-07-22
fix: not being able to remove settings and recording buttons

Kathund:
You can now remove the recording and settings buttons. They are still kept there by default but my only "issue" is that I forced my self to make an `INTERNAL` tag/prefix for naming an command. I would like to come up with a better way to do this but I personally couldn't come up with one. I am open to feedback on how we implement start video recording then


src/configuration/modules/config/menus/dashboard/index.ts
```diff
@@ -1,3 +1,4 @@
+import { getRecordingPath, isRecording } from 'src/components/menus/dashboard/shortcuts/helpers';
```
Jas-SinghFSU, line 1: This is going to cause a circular import issue. Options is importing these helpers, but these helpers (code in that file) are importing options. So you have a circular import reference which prevents the application from starting up.

Jas-SinghFSU:
```ts
const ShortcutButton = ({ shortcut, ...props }: ShortcutButtonProps): JSX.Element => {
    const isRecordingButton = shortcut.command.get() === 'hyprpanel_record';

    return (
        <button
            vexpand
            tooltipText={shortcut.tooltip.get()}
            className={
                isRecordingButton
                    ? bind(isRecording).as((rec) => `${props.className || ''} ${rec ? 'active' : ''}`.trim())
                    : props.className
            }
            onButtonPressEvent={(_, event) => {
                const buttonClicked = event.get_button()[1];

                if (buttonClicked !== Gdk.BUTTON_PRIMARY) {
                    return;
                }

                if (!isRecordingButton) {
                    handleClick(shortcut.command.get());
                    return;
                }

                const sanitizedPath = getRecordingPath().replace(/"/g, '\\"');

                if (isRecording.get() === true) {
                    App.get_window('dashboardmenu')?.set_visible(false);
                    const command = `${SRC_DIR}/scripts/screen_record.sh stop "${sanitizedPath}"`;
                    executeCommand(command);
                    return;
                }

                const monitorDropdownList = MonitorListDropdown() as Gtk.Menu;
                monitorDropdownList.popup_at_pointer(event);
            }}
        >
            <label className={'button-label txt-icon'} label={shortcut.icon.get()} />
        </button>
    );
};
```
I've updated the recording button logic for `ShortcutButton`.

src/configuration/modules/config/menus/dashboard/index.ts
```diff
@@ -53,11 +54,25 @@ export default {
...
                 command: opt(`bash -c "${SRC_DIR}/scripts/snapshot.sh"`),
             },
+            shortcut4: {
+                icon: opt('󰑊'),
+                tooltip: opt('Record Screen'),
+                command: opt(
+                    isRecording.get() === true
+                        ? `${SRC_DIR}/scripts/screen_record.sh stop "${getRecordingPath().replace(/"/g, '\\"')}"`
+                        : 'INTERNAL_RECORDING_START',
+                ),
```
Jas-SinghFSU, line 74:
```suggestion
                command: opt('hyprpanel_record'),
```
Let's keep this simple. The component will handle the stopping, starting, etc.

Jas-SinghFSU:
That should cover it. Great work and thank you for contributing <3

Jas-SinghFSU:
Hey @Kathund, thanks for this PR and for the time you put into it.

Thanks again! <3

Issue #1116, opened 2025-10-12
Make laptop-specific packages optional


bitwise74:
As it stands right now, the `bluez-utils`, `bluez` and `brightnessctl` packages are required to run hyprpanel otherwise it won't launch. These packages are superfluous for users that don't have either capability on their system (eg. PC users with Bluetooth). It would be more appropriate for these packages be optional, only required for users who intend to use these features.

If this issue is recognized I'm able to create a pull request with the edited code to make this real

PR #1119, opened 2025-10-14
Make laptop specific packages optional

bitwise74:
This pull request is directly related to issue #1116. It makes the `bluez` and `bluez-utils` packages optional and improves feedback for when a user wants to use these bar components with missing packages

New bluetooth component:
image
The label now explicitly states "Unavailable" instead of just saying "Off" and clicking on the element lets the user know the bluetooth service is missing

New power menu:
image
In case brightnessctl is missing, the slider gets replaced with a message saying it wasn't found on the machine

### DIFF
--- a/src/components/bar/modules/clock/index.tsx
+++ b/src/components/bar/modules/clock/index.tsx
@@ -11,7 +11,7 @@ import { throttledScrollHandler } from '../../utils/input/throttle';
 const { format, icon, showIcon, showTime, rightClick, middleClick, scrollUp, scrollDown } = options.bar.clock;
 const { style } = options.theme.bar.buttons;
 
-const time = Variable.derive([systemTime, format], (c, f) => c.format(f) ?? '');
+const time = Variable.derive([systemTime, format], (c, f) => c.format(f.replace(/\\n/g, '\n')) ?? '');
 
 const Clock = (): BarBoxChild => {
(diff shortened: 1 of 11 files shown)


src/core/initialization/index.ts
```diff
@@ -56,11 +57,22 @@ export class InitializationService {
      */
     private static async _initializeStartupScripts(): Promise<void> {
         try {
-            execAsync(`python3 ${SRC_DIR}/scripts/bluetooth.py`).catch((err) =>
-                console.error('Failed to initialize bluetooth script:', err),
-            );
+            const bluetoothStatus = SystemUtilities.checkServiceStatus(['bluetooth.service']);
+
+            if (bluetoothStatus === 'DISABLED') {
```
Jas-SinghFSU: Is there any harm in enabling the script even when bluetooth service is disabled? In case it gets enabled while hyprpanel is running? Or would it require a restart anyways?

bitwise74: I messed around with the code a bit and the only drawback I could find is hyprpanel throwing an error on startup if bluetooth is disabled/missing but I suppose that's not too bad because it runs just well. I'll roll the change back

Jas-SinghFSU:
Seems like there are some linting errors. Could you run `npm run lint:fix`?

PR #1123, opened 2025-11-01
Add option to hide battery widget if no battery is present

### LINKED ISSUE #940: dont show battery on 0%
**Is your feature request related to a problem? Please describe.**
On laptop it is fine, but on a desktop pc there is a permanent 0% battery icon since there is no battery supply

**Describe the solution you'd like**
Provide and option to not show battery at 0%

ranger-ross:
This PR adds a "Hide When No Battery Found" option to Battery section of the Bar configuration.

image

When its enabled the battery module will be hidden if the battery type is not `ASTAL_BATTERY_TYPE_BATTERY`. (see available types [here](https://aylur.github.io/libastal/battery/enum.Type.html)).

This option is disabled by default.

| Disabled | Enabled |
|--------|--------|
| image | image |

fixes #940


ranger-ross:
I originally tried that but there were two problems I ran into:
1. It seems `isVisible` does not respect state changes and needs to be a primative `boolean`. 
2.  Even when hardcoding is `isVisible` to `false` there is still an empty container that is rendered by the `WidgetContainer`

I could be incorrect since I've never used this framework but I couldn't get it to work.

Jas-SinghFSU:
Hey @ranger-ross, thanks for this PR and for the time you put into it.

Thanks again! <3

Issue #1128, opened 2025-11-09
Matugen theme not applied


DaveBeat64:
**Describe the bug**
Enabling matugen theming does nothing
**To Reproduce**
Install latest ags-hyprpanel-git from aur along with all dependencies
Set a wallpaper in hyprpanel and enable matugen theming
Observe that nothing changes
**Expected behavior**
The hyprpanel theme should update based on the selected wallpaper

**Desktop:**
 - Distribution: Arch Linux
 - Window Manager/Desktop Environment: Hyprland

zapds:
downgrading to matugen version 2.4.1 worked for me.

ZODIAC3K:
Yes I am facing the same issue on Arch with the latest "ags-hyprpanel-git". I will downgrade to 2.4.1 and check, then update here in a few hours.

zapds:
@ZODIAC3K I installed using cargo

zapds:
cargo install matugen --version 2.4.1 
@ZODIAC3K Did you try this?

ZODIAC3K:
no i didn't i fixed by forking it

ZODIAC3K:
Please close this issue as this has been resolved and merged 😇.

PR #1130, opened 2025-11-10
Fix Matugen Color Extraction from Mode-Based Objects

ZODIAC3K:
## Issue: #1128 

Matugen was returning color objects with mode-based properties (`{dark, default, light}`) instead of hex strings when using `--json hex` flag. The `getMatugenHex()` function expected hex strings but received objects, causing errors like:

```
[Matugen] Invalid color properties (not hex colors): background (value: [object Object]), ...
```

The function `getMatugenVariations()` tried to access properties like `matugenColors.primary` expecting a hex string, but got an object instead.

## Solution

Added destructuring logic in `generateMatugenColors()` to extract hex values from mode-based color objects:

- Extracts the correct color based on the selected mode (`dark` or `light`)
- Falls back to `default` if mode-specific color is not available
- Handles edge cases (direct hex property, RGB values)
- Added comprehensive comments documenting the structure transformation

### `src/services/matugen/index.ts`

- Added color object destructuring logic to extract hex values based on mode
- Added detailed comments documenting the structure before/after destructuring
- Removed excessive validation code
- Added comment in `getMatugenHex()` clarifying it expects destructured hex strings

### `env.d.ts`

- Added TypeScript declaration for GIF file imports (`*.gif`)

### `src/style/optionsTrackers.ts`

- Fixed TypeScript error for `connect()` method on `WallpaperService` by adding type assertion to `GObject.Object`

### `.gitignore`

- Added `pnpm-lock.yaml` to ignore list

## Testing

- ✅ Matugen colors are correctly extracted from mode-based objects
- ✅ Bar colors update correctly when wallpaper changes
- ✅ Matugen works for both dark and light modes

## Example

**Before (what matugen returns):**

```json
{
    "colors": {
        "primary": { "dark": "#4dd0a6", "default": "#4dd0a6", "light": "#006c4c" },
        "background": { "dark": "#0f1512", "default": "#0f1512", "light": "#f5fbf6" }
    }
}
```

**After (what we pass to getMatugenHex):**

```json
{
    "primary": "#4dd0a6", // extracted from obj[mode] or obj.default
    "background": "#0f1512"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,6 @@ yay -S ags-hyprpanel-git
 aylurs-gtk-shell-git
 wireplumber
 libgtop
-bluez
-bluez-utils
 networkmanager
 dart-sass
 wl-clipboard
@@ -47,6 +45,10 @@ python-gpustat
 
 ## To control screen/keyboard brightness
 brightnessctl
+
+## For bluetooth support
+bluez
+bluez-utils
 
 ## Only if a pywal hook from wallpaper changes applied through settings is desired
 pywal
--- a/src/components/bar/modules/bluetooth/index.tsx
+++ b/src/components/bar/modules/bluetooth/index.tsx
@@ -1,14 +1,16 @@
 import { Variable, bind } from 'astal';
-import AstalBluetooth from 'gi://AstalBluetooth?version=0.1';
 import { Astal } from 'astal/gtk3';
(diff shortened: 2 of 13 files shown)


regit:
Hello, I'm really enjoying hyprpanel but I'm waiting for this integration with Matugen to work. Is there anything I can do to help ?

PR #1132, opened 2025-11-14
Update bluetooth.py: fix import crash; add robust no-BT fallback

pullclone:
This commit fixes an import-time crash in bluetooth.py and makes the agent robust when DBus/GLib bindings or the BlueZ daemon are missing.

  - Pass human-readable message as the first arg and DBus error name via name= (e.g., org.bluez.Error.Rejected, org.bluez.Error.Canceled).
- Switch Agent capability to KeyboardDisplay so both PIN and passkey flows are supported.

Robust behavior when Bluetooth is unavailable
- Introduce HYPERPANEL_BT_KEEP_ALIVE env flag (default: off).
  - When deps or BlueZ are missing, either exit(0) or idle (GLib loop/sleep) based on the flag.
- Add helpers:
  - idle_or_exit to centralize graceful shutdown/idle behavior.
  - is_bluez_running to check org.bluez on the system bus before attempting registration.
- Improve user interaction and notifications:
  - request_input_sync uses zenity when available, otherwise falls back to TTY input; handles cancellations cleanly.
  - send_notification_with_actions returns a Callable[[], None] disconnect and safely unregisters signal handlers; logs and no-ops if notifications aren’t available.

Other additions
- More informative logging around agent registration and user prompts.
- Defensive handling when removing signal receivers.
- Typing polish for send_notification_with_actions.

Behavioral notes
- No behavioral change when BlueZ and deps are present: the agent registers and runs the GLib main loop as before.
- When BlueZ or deps are missing, hyprpanel no longer aborts due to a parse error; the script exits 0 by default unless keep-alive is requested.

Testing tips
- Catch syntax errors early: python3 -m compileall -q /usr/share/hyprpanel/scripts/bluetooth.py
- Simulate missing BlueZ: systemctl stop bluetooth; run the script and verify exit code 0 (or idle when HYPERPANEL_BT_KEEP_ALIVE=1).
- Start hyprpanel and confirm it no longer aborts on bluetooth.py

### DIFF
--- a/scripts/bluetooth.py
+++ b/scripts/bluetooth.py
@@ -1,173 +1,454 @@
 #!/usr/bin/env python3
+"""
+bluetooth.py: A BlueZ Agent for hyprpanel that
+gracefully no-ops if python3-dbus, python3-gi, or BlueZ aren’t present.
+"""
 
-import dbus
-import dbus.service
-import dbus.mainloop.glib
-from gi.repository import GLib
+import os
+import sys
+import time
+import shutil
 import subprocess
 import logging
+from typing import Optional, Callable
 
-logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+# -------------------------------------------------------------------
+# Logger setup
+# -------------------------------------------------------------------
+logging.basicConfig(
+    level=logging.INFO,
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Hey @pullclone, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #1166, opened 2026-01-05
allow different colors for CPU based on load

Dieterbe:


Jas-SinghFSU:
Closing because of Wayle.

PR #1173, opened 2026-01-08
feat(hyprsunset): Add option to send no temperature

buraksirma:
For users with time-based temperature configurations in hyprsunset.conf, sending an explicit temperature value overrides the configuration file settings. This PR adds a 'none' option that allows hyprsunset to use its default configuration without override.


Jas-SinghFSU:
Hey @buraksirma, thanks for this PR and for the time you put into it.

Thanks again! <3

PR #1174, opened 2026-01-11
chore: replace deprecated swww command with the direct successor awww

RodoMa92:
This is needed since the old dependency has been deprecated some times ago and the theme extractor is not working correctly with the new package name.


Jas-SinghFSU:
Hey @RodoMa92, thanks for this PR and for the time you put into it.

Thanks again! <3

Issue #1185, opened 2026-03-22
swww is getting renamed to awww


PerchunPak:
In the newest release, swww binary will be renamed to awww. I would expect hyprpanel to break beacuse of this, but did not test it

https://codeberg.org/LGFae/awww/releases/tag/v0.12.0

AquilaIrreale:
Can confirm, it does

tobias-haenel:
potential hotfix for the warning (and probably the feature for setting wallpapers through hyprpanel):

symlinking `awww` to `swww` can work as long as the cli doesn't change.
`sudo ln -s /usr/bin/awww /usr/local/bin/swww`
`sudo ln -s /usr/bin/awww-daemon /usr/local/bin/swww-daemon`

IlCandeDEV:
On latest version it does break.

"Can't find dependency swww" error.

Eaglesmine:
> On latest version it does break.
> 
> "Can't find dependency swww" error.

I found a fix. Just add into hyprland.conf "exec-once = awww-daemon"
And make this:

> symlinking `awww` to `swww` can work as long as the cli doesn't change. 
`sudo ln -s /usr/bin/awww /usr/local/bin/swww`
`sudo ln -s /usr/bin/awww-daemon /usr/local/bin/swww-daemon`

IlCandeDEV:
Works for me, but the fix in the hyprpanel code should be pretty easy to implement. There is already a pull request in evaluation.

PR #1187, opened 2026-03-27
Swww to awww migration

mvioole:
# Problem 
After swww had been renamed to awww, hyprpanel wasn't able to set up a wallpaper because the referenced swww command doesn't exist anymore

# Solution
As stated by swww's creator in [here](https://www.lgfae.com/posts/2025-10-29-RenamingSwww.html) awww is interchangeable with swww, therefore to fix the problem it is sufficient to change all occurrencies of "swww" to "awww". All classes and function name related to this service were renamed as well just for clarity

---
The solution is working fine and no noticeable bug has been found on my end


src/services/wallpaper/AwwwDaemon.ts
```diff
@@ -41,35 +41,35 @@ export class SwwwDaemon {
...
     }
 
     /**
-     * Checks if the swww daemon is currently running
+     * Checks if the awww daemon is currently running
      */
     private async _checkIfRunning(): Promise<boolean> {
         try {
-            await execAsync('swww query');
+            await execAsync('awww query');
```
PerchunPak, line 60: Ideally, this should check if `swww` is available, and if not, fall back to the `awww` binary

package-lock.json
PerchunPak, line 1: Please rollback all of the changes in this file (just run `git checkout master -- package-lock.json`)

PerchunPak (commented):
Also, please add `Fixes #1185` in the PR description, so people would know about this PR

Jas-SinghFSU:
Hey @mvioole , thanks for this PR and for the time you put into it.

Thanks again! <3

PR #1191, opened 2026-04-07
feat(workspaces): support workspace name keys in workspaceIconMap

### LINKED ISSUE #1190: `workspaceIconMap` support workspace names as keys
Is your feature request related to a problem? Please describe.
I'm always frustrated when icons in the workspace bar swap between reboots. I use named special workspaces (special:discord, special:spotify, special:steam) and want to assign custom icons to each one via workspaceIconMap. This is currently impossible to do reliably because Hyprland assigns special workspaces negative numeric IDs based on creation order at runtime — whichever app opens first gets -98, the next gets -97, and so on. The order is non-deterministic, so the icon mapping is wrong after most reboots.

Describe the solution you'd like
Allow workspaceIconMap to accept workspace names as keys in addition to numeric IDs:
  "bar.workspaces.workspaceIconMap": {
    "special:discord": { "icon": "󰙯", "color": "#7289da" },
    "special:spotify": { "icon": "󰓇", "color": "#1db954" },
    "special:steam":   { "icon": "󰓓", "color": "#c6d4df" }
  }
Named keys should take priority over numeric keys when both could match. Existing numeric key behaviour should be unchanged.

Describe alternatives you've considered
- Forcing consistent workspace creation order at startup via a script — doesn't work because Hyprland destroys empty special workspaces before apps finish loading
- Using showApplicationIcons: true — shows correct icons but removes workspace numbers from regular workspaces, which is an unwanted trade-off

Additional context
This affects anyone using named special workspaces, which is a common Hyprland pattern for persistent app drawers (Discord, Spotify, Steam etc). The feature would make workspaceIconMap actually usable for this use case.

AlexTrebs:
Special workspaces in Hyprland receive non-deterministic negative IDs based on creation order at runtime, making numeric key mappings unreliable across reboots. This adds support for named keys (e.g. "special:discord") in workspaceIconMap, with name-based lookup taking priority over numeric ID lookup as a fallback.

Closes #1190

### DIFF
--- a/src/components/bar/modules/workspaces/helpers/index.ts
+++ b/src/components/bar/modules/workspaces/helpers/index.ts
@@ -38,8 +38,8 @@ const isWorkspaceActiveOnMonitor = (monitor: number, i: number): boolean => {
  *
  * @returns The icon for the workspace as a string. If no icon is found, returns the workspace index as a string.
  */
-const getWsIcon = (wsIconMap: WorkspaceIconMap, i: number): string => {
-    const iconEntry = wsIconMap[i];
+const getWsIcon = (wsIconMap: WorkspaceIconMap, i: number, wsName?: string): string => {
+    const iconEntry = (wsName !== undefined ? wsIconMap[wsName] : undefined) ?? wsIconMap[i];
     const defaultIcon = `${i}`;
 
     if (iconEntry === undefined) {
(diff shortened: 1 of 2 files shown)


Jas-SinghFSU:
Hey @AlexTrebs , thanks for this PR and for the time you put into it.

Thanks again! <3

